PR #144, opened 2023-06-24
Refactoring `dyn-abi` to performance parity with ethabi

prestwich:
This PR brings dyn-abi to performance parity with ethabi (or better) on all benchmarks

It does this by

- removing unnecessary type checks during value encoding
- removing buffer copies during decoding into tokens
- bypassing tokenization during value encoding

Followup work may bypass decoding into tokens and decode directly into values

## Motivation

dyn-abi is slow and should be fast 😤 

## Solution

make it faster

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/dyn-abi/benches/abi.rs
+++ b/crates/dyn-abi/benches/abi.rs
@@ -54,27 +54,15 @@ fn dyn_abi_encode(c: &mut Criterion) {
     g.bench_function("single", |b| {
         let input = encode_single_input();
         b.iter(|| {
-            let ty = DynSolType::String;
             let value = DynSolValue::String(input.clone());
-            ty.encode_single(black_box(&value))
+            black_box(&value).encode()
         });
     });
 
     g.bench_function("struct", |b| {
-        let tys = vec![
-            DynSolType::Address,
-            DynSolType::Address,
-            DynSolType::Uint(24),
-            DynSolType::Address,
-            DynSolType::Uint(256),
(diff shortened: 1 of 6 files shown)


crates/dyn-abi/benches/abi.rs
```diff
@@ -54,27 +54,15 @@ fn dyn_abi_encode(c: &mut Criterion) {
...
-            DynSolType::Address,
-            DynSolType::Uint(256),
-            DynSolType::Uint(256),
-            DynSolType::Uint(256),
-            DynSolType::Uint(160),
-        ];
-        let ty = DynSolType::Tuple(tys);
         let input = encode_struct_sol_values();
         let input = DynSolValue::Tuple(input.to_vec());
-        b.iter(|| ty.encode_single(black_box(&input).clone()));
```
prestwich, line 77: this was doing an extra alloc in the encode_single

DaniPopes (approved):

DaniPopes:
Can you share the benchmark results (with % gains)?

prestwich:
Here's the comparison to main. 

A significant % of gains were from making the benches apples-to-apples. The versions in main actually cause the dyn-abi to do extra work that ethabi was not required to. I'd guess that about 100ns of the improvement is from making the bench equivalent

Screenshot 2023-06-26 at 8 59 37 AM

PR #149, opened 2023-06-25
Format and comment fixes

Kuly14:
Very minor pr that fixes a weird format error and 2 comments.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,6 +10,8 @@ The contributors guide includes instructions for running rustfmt and building th
 documentation.
 -->
 
+<!-- ** Please select "Allow edits from maintainers" in the PR Options ** -->
+
 ## Motivation
 
 <!--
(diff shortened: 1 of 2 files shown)


crates/primitives/src/bits/macros.rs
```diff
@@ -173,7 +173,7 @@ macro_rules! wrap_fixed_bytes {
             /// Returns the inner bytes array.
             #[inline]
             pub const fn into_array(self) -> [u8; $n] {
-                self.0 .0
+                self.0.0
```
DaniPopes, line 176: This is actually intentional since Rust parses the `0.0` as a float literal and rustfmt will prefer the separated version, but both work

Kuly14: Oh ok, had no idea. Thanks for explaining!

DaniPopes: rustfmt formats `self.0.0` to `self.0 .0` if you try it (top left, tools -> rustfmt):
<https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fc90885df6ae8d4805352d11061f9a2b>

DaniPopes (approved):

prestwich:
this work is now in #151 because the rebase did something weird to github's interpretation of the branch and i can't reopen this. 😮‍💨

PR #150, opened 2023-06-25
chore: clean ups

DaniPopes:
## Solution

- d6fb51dd9947e243525117189f8f9d683f490598: random stuff i found
- c43ac0532ad8ab62e4fc5397aececf15ad8cf20d: rename `no_std_prelude` to `private` and export some more items
- 5228cf70fb737ae244825ae73143f9a18d89b5b0: refactor `eip712_domain` to use only 1 arm with the question mark repetition `$()?`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/sol-types/src/eip712.rs
DaniPopes, line 1: this is just swapping `Address` with `Bool`

crates/sol-types/src/types/data_type.rs
DaniPopes, line 1: The diff is pretty horrendous here, but it's just swapping `Address` with `Bool`

prestwich (approved):
LG2M

PR #153, opened 2023-06-26
feat: `SolEnum` and `SolInterface`

### LINKED ISSUE #37: Add `sol!` support for enums
Should `#[repr(u8)]`, add convert functions and impl `SolType` (new `SolEnum` trait?)

### LINKED ISSUE #106: [Feature] Convenient way to decode a contract error
### Component

sol-type, sol! macro

Currently we generate a `ContractErrors` enum, which is fine for custom errors but cannot be used to also decode Solidity panics and reverts.
We should provide some way to easily decode all of the above in one method

### Additional context

_No response_

DaniPopes:
## Motivation

Closes #37 
Closes #106 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/sol-types/src/types/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Contains a [`Revert`] or [`Panic`] error, or a custom error.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ContractError<T> {
+    /// A contract's custom error.
+    CustomError(T),
+    /// A generic revert. See [`Revert`] for more information.
+    Revert(Revert),
+    /// A panic. See [`Panic`] for more information.
+    Panic(Panic),
+}
```
prestwich, line 74: Should there be an empty variant?

crates/sol-types/src/types/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// [`SolCall`]: crate::SolCall
+/// [`SolError`]: crate::SolError
+///
+/// ### Implementer's Guide
+///
+/// We do not recommend implementing this trait directly. Instead, we recommend
+/// using the [`sol`][crate::sol] proc macro to parse a Solidity contract
+/// definition.
+pub trait SolCalls: Sized {
```
prestwich, line 15: This is something like `SolInterface` no? It's an enum over all expected inputs/outputs

crates/sol-macro/src/expand/contract.rs
```diff
@@ -120,64 +120,75 @@ fn expand_call_like_enum(
     assert_eq!(variants.len(), types.len());
     let name_s = name.to_string();
     let count = variants.len();
-    let min_data_len = min_data_len.min(4);
     quote! {
         #(#attrs)*
         pub enum #name {
             #(#variants(#types),)*
```
prestwich, line 126: should there be a variant for `Bare` calls, e.g. fallback/receive?

If so, that variant should contain a `Vec<u8>` containing the calldata

We can distinguish between `Bare` and `Unknown` by checking lenght. `len() % 32 == 4` is Unknown and should error. `len % 32 == 0` is `Bare`

crates/sol-types/src/types/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// The name of this type.
+    const NAME: &'static str;
+
+    /// The minimum length of the data for this type.
+    ///
+    /// This does *not* include the selector's length (4).
+    const MIN_DATA_LENGTH: usize;
```
prestwich, line 22: this doesn't work in the presence of a fallback/receive

crates/sol-types/src/types/enum.rs
```diff
@@ -0,0 +1,36 @@
...
+use alloc::vec::Vec;
+
+/// Solidity enum. This is always a wrapper around a [`u8`].
+///
+/// ### Implementer's Guide
+///
+/// We do not recommend implementing this trait directly. Instead, we recommend
+/// using the [`sol`][crate::sol] proc macro to parse a Solidity error
+/// definition.
+pub trait SolEnum: Sized + Copy + Into<u8> + TryFrom<u8, Error = crate::Error> {
```
prestwich, line 11: enums should `impl SolType` yes?

crates/sol-types/src/types/calls.rs
prestwich, line 15: should we enable iterating over variant names/signatures/selectors?

crates/sol-types/src/types/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+/// [`SolCall`]: crate::SolCall
+/// [`SolError`]: crate::SolError
+///
+/// ### Implementer's Guide
+///
+/// We do not recommend implementing this trait directly. Instead, we recommend
+/// using the [`sol`][crate::sol] proc macro to parse a Solidity contract
+/// definition.
+pub trait SolCalls: Sized {
+    /// The name of this type.
```
prestwich, line 16: this is the rust typename of the enum, yes?

crates/sol-macro/src/expand/contract.rs
```diff
@@ -120,64 +120,75 @@ fn expand_call_like_enum(
...
-                    let selector: &[u8; 4] =
-                        ::core::convert::TryInto::try_into(selector).expect("unreachable");
-                    match *selector {
-                        #(<#types as ::alloy_sol_types::#trt>::SELECTOR => {
-                            return <#types as ::alloy_sol_types::#trt>::decode_raw(data, validate)
-                                .map(Self::#variants)
-                        })*
-                        _ => {}
-                    }
+        impl ::alloy_sol_types::SolCalls for #name {
```
prestwich, line 130: can we also implement a fallible downcast for each type? `fn as_#variant(&self) -> Option<&#variant_ty>`

crates/sol-types/src/types/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// A generic revert. See [`Revert`] for more information.
+    Revert(Revert),
+    /// A panic. See [`Panic`] for more information.
+    Panic(Panic),
+}
+
+impl<T: SolCalls> SolCalls for ContractError<T> {
+    const NAME: &'static str = "ContractError";
+
+    // revert is 64, panic is 32
```
prestwich, line 79: bare reverts are still common in widely deployed pre-panic contracts

crates/sol-types/src/types/calls.rs
DaniPopes: zero?

crates/sol-types/src/types/calls.rs
DaniPopes: yes

crates/sol-types/src/types/calls.rs
DaniPopes: revert here is the revert struct which is `error(string)`

crates/sol-macro/src/expand/contract.rs
DaniPopes: sure

crates/sol-types/src/types/calls.rs
DaniPopes: Makes sense

crates/sol-types/src/types/calls.rs
DaniPopes: yeah but it's implemented for both calls and custom errors, doesn't have a name in abigen either, other than the struct named `<contract_name>{Calls,Errors}`

crates/sol-macro/src/expand/contract.rs
DaniPopes: I think this can be done in another PR, since we don't handle any type of function (constructor, fallback...) other than literally `function`

crates/sol-types/src/types/enum.rs
DaniPopes: Should they? They're always u8

crates/sol-types/src/types/calls.rs
DaniPopes: I initially had a `const SELECTORS: &'static [[u8; 4]]` but that doesn't work with the nested `ContractError` I later added

crates/sol-types/src/types/calls.rs
prestwich: so it's 0 for any contract that has a fallback/receive function? or is it None for those contracts and Some(_) for other contracts? related to #157

crates/sol-types/src/types/enum.rs
prestwich: sure, but we can include an enum in a solidity struct, or in function args. so it should be a SolType

crates/sol-macro/src/expand/contract.rs
prestwich: strum does provide `fn is_#variant(&self) -> bool` tho 🤔

crates/sol-macro/src/expand/mod.rs
```diff
@@ -109,33 +113,32 @@ impl<'ast> ExpCtxt<'ast> {
...
+                let Some(resolved) = map.get(name.last_tmp()) else { return };
+                ty.clone_from(resolved);
+                i += 1;
+            });
+            if i >= RESOLVE_LIMIT {
+                let msg = "\
+                    failed to resolve types.\n\
+                    This is likely due to an infinitely recursive type definition.\n\
+                    If you believe this is a bug, please file an issue at \
+                    https://github.com/ethers-rs/core/issues/new/choose";
```
onbjerg, line 137:
```suggestion
                    https://github.com/alloy-rs/core/issues/new/choose";
```

crates/sol-types/tests/ui/type.stderr
```diff
@@ -57,3 +57,106 @@ error: intX must be a multiple of 8 up to 256
     |
 199 |         int7 a;
     |         ^^^^
+
+error: failed to resolve types.
+       This is likely due to an infinitely recursive type definition.
+       If you believe this is a bug, please file an issue at https://github.com/ethers-rs/core/issues/new/choose
```
onbjerg, line 63:
```suggestion
       If you believe this is a bug, please file an issue at https://github.com/alloy-rs/core/issues/new/choose
```

crates/sol-types/src/types/enum.rs
DaniPopes: works fine up until `fn detokenize` because u8 -> Self is fallible:

```rust
impl Encodable<#name> for #name {...}
impl SolType for #name {
    type RustType = #name;

    ...

    fn detokenize(token: Self::TokenType<'_>) -> Self::RustType {
        let x = #uint8_st::detokenize(token);
        todo!("???")
    }
}
```

prestwich: the behavior for `detokenize` an invalid value into a u8 is to ignore all high bytes. should the behavior for an enum be similar? 🤔

DaniPopes: Maybe with an `__Invalid = u8::MAX` variant for default?

crates/sol-types/src/types/calls.rs
DaniPopes: This just kinda breaks the entire thing, since now we can't assume that data has a selector. I'm not sure how to do this, I think it's better we don't include this

prestwich: yeah, I guess the question is whether we want this trait/enum to capture:
1. ALL inputs and outputs (known ABI & unknown ABI & Empty)
2. NON-EMPTY inputs and outputs (known ABI & unknown ABI)
3. KNOWN inputs and outputs (known ABI & Empty)
4. KNOWN & NON-EMPTY inputs and outputs (known ABI only)

Right now we have option 4. Do we have a plan for capturing the remainder?

crates/sol-types/src/types/enum.rs
prestwich: That sounds right to me. and then type_check fails for invalid

crates/sol-types/src/types/calls.rs
DaniPopes: I'm thinking 4 is fine for now, maybe 3 in the future. I don't want to overload this PR

crates/sol-types/src/types/calls.rs
DaniPopes: See https://github.com/alloy-rs/core/pull/153#discussion_r1244051343

crates/sol-macro/src/expand/contract.rs
DaniPopes: I've added them anyway, `From`, `TryInto`, `is_*`, and `as_*{,_mut}`. Still not sure about this, maybe we conditionally add them with attributes later (i'm still tinkering with the `#[sol(...)]` attributes)

crates/sol-types/src/types/calls.rs
DaniPopes: What do you think of
```rust
trait {
  fn selectors() -> Selectors<Self> {}
  fn selector_at(i: usize) -> Option<[u8; 4]> {}
}

struct Selectors<T> { index: usize }
impl<T> Iterator for Selectors<T> { ... }
```

DaniPopes: I've addded the above interface, as well as re-adding the `const SELECTORS` as a associated constant on the enum itself, instead of on the trait

crates/sol-macro/src/expand/contract.rs
```diff
@@ -136,14 +162,33 @@ impl CallLikeExpander {
...
-                .map(|error| r#type::params_base_data_size(cx, &error.parameters))
+                .map(|error| ty::params_base_data_size(cx, &error.parameters))
                 .min()
                 .unwrap(),
             trait_: Ident::new("SolError", Span::call_site()),
             data: CallLikeExpanderData::Error { selectors },
         }
     }
 
+    /*
```
prestwich, line 173: what's the status of this code?

crates/sol-macro/src/expand/contract.rs
```diff
@@ -56,13 +59,15 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
 
-    // TODO: events enum
-    // let events_enum = (events.len() > 1).then(|| {
-    //     let mut attrs = d_attrs;
-    //     let doc_str = format!("Container for all the `{name}` events.");
-    //     attrs.push(parse_quote!(#[doc = #doc_str]));
-    //     CallLikeExpander::from_events(cx, name, events).expand(&attrs)
-    // });
+    // TODO
+    /*
```
prestwich, line 63: what's the status of this code?

crates/sol-types/src/types/interface.rs
```diff
@@ -87,7 +90,7 @@ pub enum ContractError<T> {
     Panic(Panic),
 }
 
-impl<T: SolCalls> SolCalls for ContractError<T> {
+impl<T: SolInterface> SolInterface for ContractError<T> {
```
prestwich, line 93: this allows accidentally mixing calls and errors, yeah?

`ContractError<MyContractCalls>` is valid, as is `ContractError<MyContractErrors>`

is there a way to fix this in the type system?

DaniPopes: Would either require a generic or assoc type which implements `Sol*`, but idk.
I'd also like to not have different traits for the event enum with something like `type Selector = [u8; _]` but not sure if doable.
I'd leave this for later since it's not that big of a deal.

crates/sol-macro/src/expand/contract.rs
DaniPopes: refactored to expand the enum but no SolInterface trait impl

crates/sol-types/src/types/calls.rs
prestwich: I like it.

crates/sol-types/src/types/interface.rs
prestwich: should we have a trait to capture this somehow? 

```
trait SelectorPrefixed {}
trait SolError: SelectorPrefixed {}
```
that sort of thing?

crates/sol-macro/src/expand/contract.rs
```diff
@@ -64,124 +74,339 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
+/// impl SolInterface for #name {
+///     ...
+/// }
+///
+/// impl #name {
+///     #(
+///         pub fn #is_variant,#as_variant,#as_variant_mut(...) -> ... { ... }
+///     )*
+/// }
+/// ```
```
prestwich, line 104: this is a good refactor 😍

crates/sol-macro/src/expand/contract.rs
```diff
@@ -64,124 +74,339 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
+                        Self::#variants(inner) =>
+                            <#types as ::alloy_sol_types::#trait_>::encode_raw(inner, out),
+                    )*}
+                }
+            }
+        }
+    }
+
+    fn expand_event(self, attrs: &[Attribute]) -> TokenStream {
+        // TODO: SolInterface for events
```
prestwich, line 282: capture in an issue as this needs some conceptualizing I think? 🤔

prestwich (approved):
this is huge. we should get it in and then capture remaining work items as issues

PR #164, opened 2023-07-02
feat(json-abi): add more impls

DaniPopes:
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/sol-macro/src/expand/mod.rs
```diff
@@ -129,8 +129,12 @@ impl<'ast> ExpCtxt<'ast> {
                     return
                 }
                 let ty @ Type::Custom(_) = ty else { return };
-                let Type::Custom(name) = &*ty else { unreachable!() };
-                let Some(resolved) = map.get(name.last_tmp()) else { return };
+                let Type::Custom(name) = &*ty else {
```
prestwich, line 132: fmt drive-by

crates/syn-solidity/src/ident/path.rs
```diff
@@ -129,7 +129,9 @@ impl SolPath {
     }
 
     pub fn span(&self) -> Span {
-        let Some(first) = self.0.first() else { return Span::call_site() };
+        let Some(first) = self.0.first() else {
```
prestwich, line 132: fmt drive-by

prestwich (approved):

PR #173, opened 2023-07-03
feat(sol-macro): `#[sol]` attributes and JSON ABI support

### LINKED ISSUE #73: Support generating `sol!` from ABI JSON
Needs #72.

- Add inner attributes to `File`
  - Parse `#[sol(...)]` attributes, maybe `#![sol(include = "path/to/abi.json")]`
- Parse included JSON string as dynamic JSON ABI
- Generate a Solidity interface string
- Parse Solidity interface string into a `TokenStream`, and call `sol`

### LINKED ISSUE #143: [Feature] Document known syn-solidity limitations
### Component

syn-solidity

Listed in the description of #119

### Additional context

_No response_

### LINKED ISSUE #169: [Feature] move `sol!` doctest descriptions out of the code
### Component

sol-type, sol! macro

For rendering to markdown. Other inline comments and docs are fine.

### Additional context

E.g.

This: https://github.com/alloy-rs/core/blob/3eab410e5549cc6def0299bbb0362f1c790607ca/crates/sol-types/tests/doctests/contracts.rs#L5-L10

To here: https://github.com/alloy-rs/core/blob/3eab410e5549cc6def0299bbb0362f1c790607ca/crates/sol-macro/src/lib.rs#L62-L63

DaniPopes:
## Motivation

Closes #73
Closes #143
Closes #169

## Solution

See `sol!` docs

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/dyn-abi/src/value.rs
```diff
@@ -654,6 +654,7 @@ impl DynSolValue {
     /// # Examples
     ///
     /// ```ignore (pseudo-code)
+    ///
```
prestwich, line 657: this will produce a blank line in the docs, no?

crates/sol-types/tests/ui/type.rs
```diff
@@ -202,7 +202,7 @@ sol! {
 
 sol! {
     struct CustomTypes {
-        bytes_ a;
+        // bytes_ a; inconsistent output
```
prestwich, line 205: what's going on here?

crates/sol-types/tests/doctests/json.rs
```diff
@@ -0,0 +1,45 @@
+use alloy_sol_types::{sol, SolCall};
+
+sol!(
+    MyJsonContract1,
+    r#"[
```
prestwich, line 5: add a test case one for a filepath?

crates/sol-types/Cargo.toml
```diff
@@ -35,4 +35,5 @@ trybuild = "1.0"
 [features]
 default = ["std"]
 std = ["alloy-primitives/std", "hex/std", "serde?/std"]
+sol-json = ["alloy-sol-macro/json"]
```
prestwich, line 38: should this be a default feature?

crates/sol-macro/src/lib.rs
```diff
@@ -17,61 +19,164 @@
...
 /// [ABI]: https://docs.soliditylang.org/en/latest/abi-spec.html
 /// [EIP-712]: https://eips.ethereum.org/EIPS/eip-712
-/// [`alloy-sol-types`]: https://docs.rs/alloy-sol-types
 ///
 /// # Examples
 ///
-/// Note: the following examples cannot be tested here because the generated
-/// code references `alloy-sol-types`, so they are [tested in that crate]
-/// [tests] and included with `include_str!` in this doc instead.
+/// > Note: the following example code blocks cannot be tested here because the
```
prestwich, line 40: good change moving these in

prestwich (commented):
partial review

crates/sol-types/tests/ui/type.rs
DaniPopes: `cargo t` vs `cargo t --workspace` have different output for the recursive error below, this one's fine

crates/sol-types/tests/doctests/json.rs
DaniPopes: added multiple

crates/sol-macro/src/expand/contract.rs
```diff
@@ -429,14 +425,5 @@ fn generate_variant_methods((variant, ty): (&Ident, &Ident)) -> TokenStream {
...
         }
-
-        #[doc = #try_into_variant_doc]
-        #[inline]
-        pub const fn #try_into_variant(self) -> ::core::result::Result<#ty, Self> {
-            match self {
-                Self::#variant(inner) => ::core::result::Result::Ok(inner),
-                _ => ::core::result::Result::Err(self),
-            }
-        }
```
DaniPopes, line 410: cannot be const if destructor cannot be evaluated, so generating this function is useless when we are already generating `TryFrom`

crates/sol-macro/src/expand/struct.rs
```diff
@@ -91,8 +91,6 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, s: &ItemStruct) -> Result<TokenStream> {
 
         #[allow(non_camel_case_types, non_snake_case, clippy::style)]
         const _: () = {
-            use ::alloy_sol_types::private::*;
```
DaniPopes, line 91: this failed with Multicall which has a `Result` struct because of import ambiguity, we don't need to do glob imports anymore

DaniPopes:
Will rewrite some logic after #181

crates/sol-types/Cargo.toml
DaniPopes: I don't think so, it does bring in quite a bit of extra deps and it should be avoided when possible

prestwich (approved):

PR #175, opened 2023-07-04
feat(dyn-abi): add arbitrary impls and proptests

DaniPopes:
## Motivation

- impl Arbitrary (both `::arbitrary` and `proptest`) for dyn-abi enums and fuzz test them with `proptest`.
- add benchmark for `DynSolType::{parse,sol_type_name}`
- add more examples and doctests

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/dyn-abi/README.md
```diff
@@ -13,28 +13,42 @@ The dynamic encoder/decoder is significantly more expensive, especially for
...
+assert_eq!(encoded, my_data);
+# Ok::<(), Box<dyn std::error::Error>>(())
+```
+
+EIP-712:
 
-assert_eq!(decoded, my_values);
+```rust,ignore
+todo!()
 ```
```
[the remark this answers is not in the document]
prestwich: I'm still stewing on the interface, so i haven't wanted to write an example that will churn. the basic thing right now is "just deser the json"

crates/dyn-abi/src/ty.rs
```diff
@@ -79,17 +116,89 @@ pub enum DynSolType {
...
+    #[inline]
+    pub fn parse(s: &str) -> DynAbiResult<Self> {
+        TypeSpecifier::try_from(s).and_then(|t| t.resolve_basic_solidity())
+    }
+
+    /// Fallible cast to the contents of a variant.
+    #[inline]
+    pub fn as_tuple(&self) -> Option<&[Self]> {
+        match self {
+            Self::Tuple(t) => Some(t),
```
prestwich, line 157: should this use the. as_tuple macro or have similar behavior?

crates/dyn-abi/src/ty.rs
```diff
@@ -79,17 +116,89 @@ pub enum DynSolType {
...
+            Self::CustomStruct {
+                name,
+                prop_names,
+                tuple,
+            } => Some((name, prop_names, tuple)),
+            _ => None,
+        }
+    }
+
+    /// Returns wheter this type is contains a custom struct.
```
prestwich, line 177: wheter

should this function only exist if the feature is enabled?

crates/dyn-abi/src/ty.rs
```diff
@@ -79,17 +116,89 @@ pub enum DynSolType {
...
+        match self {
+            Self::Tuple(t) => Some(t),
+            _ => None,
+        }
+    }
+
+    /// Fallible cast to the contents of a variant.
+    #[inline]
+    #[allow(clippy::missing_const_for_fn)]
+    pub fn as_custom_struct(&self) -> Option<(&str, &[String], &[Self])> {
```
prestwich, line 165: should this function only exist if the feature is enabled?

crates/dyn-abi/src/ty.rs
DaniPopes: I want these to be available always, cfg is already annoying as it is with just one variant

crates/dyn-abi/src/ty.rs
DaniPopes: No I think the `as_<variant>` should only return the variant

prestwich (approved):
glad i didn't write this code 🖖

PR #182, opened 2023-07-09
feat/refactor(rlp): improve implementations

DaniPopes:
## Motivation

Refactor decode and add new impls and features.

Preferably `Decodable` would have a `'de` lifetime, and `Decodable::decode`'s param would be a `&mut Decoder<'de>` struct which would allow for easier helper method access (like `Header::decode` -> `d.decode_header()?;` etc), and deserializing borrowed byte and string slices, but this would be a significant breaking change so I didn't implement this here.

## Solution

- dedup decoding code by adding more static methods to `Header`, and by refactoring `Header::decode` to use a match expression.
- add `smol_str` dep + feature from `reth-rlp`
- add `arrayvec` impls + gate `encode_fixed_size` function to feature
- breaking changes to the free-standing `encode_*` functions' generics
- un-`#[doc(hide)]` and document `MaxEncodedLen`, `MaxEncodedLenAssoc`
- add a `Result` type alias for the decode error type
- update derive macros
  - added support for generics

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rlp/src/decode.rs
+++ b/crates/rlp/src/decode.rs
@@ -1,11 +1,10 @@
-use crate::Header;
-use bytes::{Buf, Bytes, BytesMut};
-use core::fmt;
+use crate::{Error, Header, Result};
+use bytes::{Bytes, BytesMut};
 
 /// A type that can be decoded from an RLP blob.
 pub trait Decodable: Sized {
     /// Decode the blob into the appropriate type.
-    fn decode(buf: &mut &[u8]) -> Result<Self, DecodeError>;
+    fn decode(buf: &mut &[u8]) -> Result<Self>;
 }
 
 /// An active RLP decoder, with a specific slice of a payload.
@@ -15,19 +14,14 @@ pub struct Rlp<'a> {
 
 impl<'a> Rlp<'a> {
     /// Instantiate an RLP decoder with a payload slice.
-    pub fn new(mut payload: &'a [u8]) -> Result<Self, DecodeError> {
-        let h = Header::decode(&mut payload)?;
-        if !h.list {
-            return Err(DecodeError::UnexpectedString)
-        }
-
-        let payload_view = &payload[..h.payload_length];
+    pub fn new(mut payload: &'a [u8]) -> Result<Self> {
+        let payload_view = Header::decode_bytes(&mut payload, true)?;
         Ok(Self { payload_view })
     }
 
     /// Decode the next item from the buffer.
     #[inline]
-    pub fn get_next<T: Decodable>(&mut self) -> Result<Option<T>, DecodeError> {
+    pub fn get_next<T: Decodable>(&mut self) -> Result<Option<T>> {
         if self.payload_view.is_empty() {
             Ok(None)
         } else {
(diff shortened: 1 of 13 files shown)


crates/rlp/src/lib.rs
```diff
@@ -41,3 +49,17 @@ pub const EMPTY_STRING_CODE: u8 = 0x80;
 
 /// RLP prefix byte for a 0-length array.
 pub const EMPTY_LIST_CODE: u8 = 0xC0;
+
+// Not public API.
+#[doc(hidden)]
+#[deprecated(since = "0.3.0", note = "use `Error` instead")]
```
prestwich, line 55: is this tag functional with `doc(hidden)` ?

crates/rlp/src/decode.rs
```diff
@@ -36,360 +30,168 @@ impl<'a> Rlp<'a> {
...
-        to.extend_from_slice(&from[..h.payload_length]);
-        from.advance(h.payload_length);
-
-        Self::from_utf8(to).map_err(|_| DecodeError::Custom("invalid string"))
+        Ok(vec)
     }
 }
 
-impl<E: Decodable> Decodable for alloc::vec::Vec<E> {
+#[cfg(feature = "smol_str")]
```
prestwich, line 100: does smol_str have a motivating usecase downstream? It's not really a common library

crates/rlp/src/header.rs
```diff
@@ -11,6 +12,112 @@ pub struct Header {
...
+
+            b @ EMPTY_STRING_CODE..=0xB7 => {
+                buf.advance(1);
+                payload_length = (b - EMPTY_STRING_CODE) as usize;
+                if payload_length == 1 && get_next_byte(buf)? < EMPTY_STRING_CODE {
+                    return Err(Error::NonCanonicalSingleByte)
+                }
+            }
+
+            b @ (0xB8..=0xBF | 0xF8..=0xFF) => {
```
prestwich, line 35: replace magic numbers with constants?

crates/rlp/src/lib.rs
DaniPopes: Do you mean `deprecated`? If so, yes, it worked on the other crates in this repo

prestwich: okay great 👌

crates/rlp/src/header.rs
DaniPopes: would need separate constants for the - 1 difference because `..` ranges are not allowed, at that point i don't think it's worth it

crates/rlp/src/decode.rs
DaniPopes: From reth-rlp, I'm assuming it's used somewhere there but I haven't verified that

crates/rlp/src/encode.rs
```diff
@@ -269,47 +282,53 @@ where
 }
 
 /// Calculate the length of a list.
-pub fn list_length<E, K>(v: &[K]) -> usize
+pub fn list_length<T, E>(list: &[T]) -> usize
 where
-    E: Encodable,
-    K: Borrow<E>,
+    E: ?Sized + Encodable,
+    T: Borrow<E>,
```
prestwich, line 288: nit: bound  order doesn't match nearby fns

prestwich (approved):

PR #185, opened 2023-07-12
feat: native keccak feature flag

rachel-bousfield:
# Background
As outlined in Paradigm's [introduction of Alloy](https://www.paradigm.xyz/2023/06/alloy), avoiding fragmentation of the Rust ecosystem on Ethereum means that everyone should use a consistent set of primitive types. As Rust expands into new Ethereum domains like smart contract development, generality will be key for this shared dependency.

The [`alloy_primatives`](https://docs.rs/alloy-primitives/latest/alloy_primitives/) crate included here does this very well, but could be made even better for upcoming Rust-enabled Ethereum smart contract environments like [Arbitrum Stylus](https://offchain.medium.com/hello-stylus-6b18fecc3a22). 

In addition to providing types, the crate currently pulls in `tiny_keccak`. This makes sense for most existing platforms, but in code running on top of VMs it's reasonable to accelerate common operations through so-called "host I/Os" or "traps". For example, the Stylus model includes a trap for `keccak256`.
```rs
#[link(wasm_import_module = "vm_hooks")]
extern "C" {
    fn native_keccak256(bytes: *const u8, len: usize, output: *mut u8);
}
```

Because host I/O operations run the function natively, we can get the native performance of `tiny_keccak` in a VM setting without any of the contribution to binary size. Though the example here is from Stylus, this reasoning applies to other VMs that might want to add hooks for operations like keccak. Anyone can expose a hook in their VM fulfilling `native_keccak256` above and accelerate hashing.

# Solution

This small PR introduces a new feature flag, `native-keccak`, that when enabled substitutes `tiny_keccak` for an `extern` import. Since the feature is not enabled by default, this change is 100% backward compatible.

Though the changes are to private functions, I've tried to include exhaustive documentation. Please let me know if there's anything I've missed in opening this PR and I'll be happy to make changes.

### DIFF
--- a/crates/primitives/src/utils.rs
+++ b/crates/primitives/src/utils.rs
@@ -1,6 +1,6 @@
 use crate::bits::FixedBytes;
 
-#[cfg(feature = "native-keccak")]
+#[cfg(all(feature = "native-keccak", not(feature = "tiny-keccak")))]
 #[link(wasm_import_module = "vm_hooks")]
 extern "C" {
     /// When targeting VMs with native keccak hooks, the `native-keccak` feature
@@ -22,7 +22,7 @@ extern "C" {
 ///
 /// [`keccak256`]: https://en.wikipedia.org/wiki/SHA-3
 pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> FixedBytes<32> {
-    #[cfg(not(feature = "native-keccak"))]
+    #[cfg(any(feature = "tiny-keccak", not(feature = "native-keccak")))]
     fn keccak256(bytes: &[u8]) -> FixedBytes<32> {
         use tiny_keccak::{Hasher, Keccak};
 
@@ -33,7 +33,7 @@ pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> FixedBytes<32> {
         output.into()
     }
 
-    #[cfg(feature = "native-keccak")]
+    #[cfg(all(feature = "native-keccak", not(feature = "tiny-keccak")))]
     fn keccak256(bytes: &[u8]) -> FixedBytes<32> {
         let mut output = [0u8; 32];
 
--- a/crates/primitives/src/utils.rs
+++ b/crates/primitives/src/utils.rs
@@ -5,11 +5,13 @@ use crate::bits::FixedBytes;
 extern "C" {
     /// When targeting VMs with native keccak hooks, the `native-keccak` feature
     /// can be enabled to import and use the host environment's implementation
-    /// of [`keccak256`] in place of [`tiny_keccak`].
+    /// of [`keccak256`] in place of [`tiny_keccak`]. This is overridden when
+    /// the `tiny-keccak` feature is enabled.
     ///
     /// # SAFETY
     ///
-    /// The VM accepts the preimage by pointer and length, and writes the 32-byte hash.
+    /// The VM accepts the preimage by pointer and length, and writes the
+    /// 32-byte hash.
     /// - `bytes` must point to an input buffer at least `len` long.
     /// - `output` must point to a buffer that is at least 32-bytes long.
     ///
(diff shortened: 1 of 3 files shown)


DaniPopes:
This is really nice, thank you!
Currently CI is failing because it builds with `--all-features` which would enable this feature. We can get around this by introducing a second feature for `tiny-keccak` that is enabled by default, and that takes precedence over `native-keccak256` with the `cfg_if!` like so:

```rust
pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> FixedBytes<32> {
    cfg_if::cfg_if! {
        if #[cfg(feature = "tiny-keccak")] {
            fn keccak256 ...
        } else if #[cfg(feature = "native-keccak")] {
            fn keccak256 ...
        } else {
            compile_error!("At least one Keccak-256 feature must be enabled")
        }
    }

    keccak256(bytes.as_ref())
}
```

This would also allow us to also expand this in the future with more Keccak implementations. Would you be willing to implement this?
Also, don't worry about breaking changes as we're still pre 1.0 and we'll be releasing 0.3 in a week or so.

prestwich:
@DaniPopes is this something that could/should be addressed with a "global keccak backend" abstraction? seems like overkill for now 🤔

rachel-bousfield:
I'm happy to go any route y'all think. Other options include
- changing the feature flags used in CI
- having instead a `tiny_keccak` feature enabled by default that opts _out_ of using a native keccak

DaniPopes:
We want to gate the tiny_keccak dep in case of this new feature, so you end up with 2 backends if you enable both features with `--all-features`, so yes

rachel-bousfield:
@DaniPopes edited my last comment for clarity: what I meant to say was that there's an in-between option where with no features enabled you use the native keccak hook, but replace it with `tiny_keccak` should the `tiny-keccak` feature flag be enabled.

I have no strong feelings on the mechanism and am happy to implement whichever solution y'all think is best :)

rachel-bousfield:
On further reflection we have a few issues
1. CI tests with both `--no-default-features` and `--all-features`. Hence, we can't just error if neither flag is enabled unless we're willing to modify CI.
2. If we add a `tiny_keccak` flag, then those working in a VM setting have an issue. While `alloy-primatives` by itself will work as expected, if the smart contract author then pulls in another library that depends on `alloy` with the default features enabled, suddenly everything is now using `tiny_keccak` instead of the VM hook.

I really think the solution is to have 1 flag, `native-keccak`, and to modify CI, which I'm happy to do.

rachel-bousfield:
Thanks everyone for your help with this. I've included a commit to add a `tiny-keccak` feature flag.

Importantly, neither `tiny-keccak` nor `native-keccak` are enabled by default. However, the effects of `tiny-keccak` are applied when neither are. This way `--no-default-features` and `--all-features` still work as expected.

But because neither are enabled by default, there's no worry that smart contract devs might accidentally stop using `native-keccak` by pulling in other crates that depend on Alloy.

I think it's a win-win :)

DaniPopes:
SGTM. I think it's still better we use `cfg_if` for readability, where now the first condition is `any(tiny-keccak, not(native-keccak))`. Also needs `cargo +nightly fmt`.

crates/primitives/src/utils.rs
```diff
@@ -1,17 +1,46 @@
 use crate::bits::FixedBytes;
-use tiny_keccak::{Hasher, Keccak};
+
+#[cfg(all(feature = "native-keccak", not(feature = "tiny-keccak")))]
+#[link(wasm_import_module = "vm_hooks")]
+extern "C" {
+    /// When targeting VMs with native keccak hooks, the `native-keccak` feature
```
prestwich, line 6: nit: docstring should specify that `tiny_keccak` must be disabled in order for this to work

prestwich (changes requested):

crates/primitives/src/utils.rs
rachel-bousfield: Added documentation and addressed @DaniPopes's comments :)

I had to add `cfg_if` as a new dependency btw. I went ahead and made it part of the workspace

DaniPopes (approved):

prestwich (approved):

joshuacolvin0 (commented):
LGTM

PR #199, opened 2023-07-19
feat(syn-solidity): add statements and expressions

Will-Smith11:
Cooking for another project and thought why not make a pr request. I gotta do testing and im pretty sure ive missed a few cases. However just going to keep cooking at ethcc and will do cleanup later


prestwich:
when writing rustdoc for the structs, can you reference the corresponding structure in the [solidity grammar](https://docs.soliditylang.org/en/v0.8.21/grammar.html) for each type?

Will-Smith11:
> when writing rustdoc for the structs, can you reference the corresponding structure in the [solidity grammar](https://docs.soliditylang.org/en/v0.8.21/grammar.html) for each type?

oooo like this yes

gakonst (commented):
nice - the idea here is to comptime transpile solidity to rust directly w/o jit'ing into any lower level code rigth?

Will-Smith11:
> nice - the idea here is to comptime transpile solidity to rust directly w/o jit'ing into any lower level code rigth?

yeah exactly. also a lot more user friendly to work with. I need to write some tests and stuff but super busy with work rn so the progress will be slow for the next little bit. Ik some of the Primitive devs where also interested in doing a paper based off the results not only looking at speed differences but llvm compile difference when comparing to revm.

DaniPopes:
Hey @Will-Smith11, I have some free time these coming days, do you mind if I pick this up?

Will-Smith11:
> Hey @Will-Smith11, I have some free time these coming days, do you mind if I pick this up?

go for it!

Will-Smith11:
@DaniPopes What is left to be done? I have some free time this weekend and want to push this forward?

DaniPopes:
Not much, the only thing is testing that the parsed output is what we expected, instead of just "it parses". Also operation order might not be entirely correct

Will-Smith11:
> Not much, the only thing is testing that the parsed output is what we expected, instead of just "it parses". Also operation order might not be entirely correct

:+1:

gakonst:
@Will-Smith11 would love if you PoC'ed a benchmark comparing Revm+Solidity compiled vs native transpiled with syn-solidity, Im curious how you see this working for e.g. projects like uniswap-v3

DaniPopes:
Merging for now, will open a few issues with what's missing.

DaniPopes (approved):

PR #200, opened 2023-07-20
fix: remove unused method body on solstruct

prestwich:
`SolStruct::eip712_encode_type` is incorrect and always overridden by sol macro.

We remove it

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes:
maybe we can also remove name and field constants after this?

DaniPopes (approved):

PR #202, opened 2023-07-20
feature: bytes handles numeric arrays and bytearrays in deser

### LINKED ISSUE #198: [Feature] Allow Bytes deserialization to accept arrays of `u8`
### Component

primitives

the follow json should be deserializable for `FixedBytes` and `Bytes`
```
{
   "myFixedBytes<4>": [0, 21, 3, 58]
}
```

### Additional context

_No response_

prestwich:
closes #198 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/serde.rs
+++ b/crates/primitives/src/bits/serde.rs
@@ -1,5 +1,9 @@
 use super::FixedBytes;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use alloc::fmt;
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
 
 impl<const N: usize> Serialize for FixedBytes<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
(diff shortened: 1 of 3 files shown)


crates/primitives/src/bits/serde.rs
```diff
@@ -10,13 +14,83 @@ impl<const N: usize> Serialize for FixedBytes<N> {
...
+                    N * 2
+                )
+            }
+
+            fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                <[u8; N]>::try_from(v)
+                    .map(FixedBytes)
```
DaniPopes, line 36: maybe this should be done in hex? https://docs.rs/const-hex/latest/src/const_hex/serde.rs.html#60-97

prestwich: mmm. I didn't want to introduce non-hex behavior to the `const_hex` crate, which this does

This would be a good chance to reify our serialization strategy for these things. Something like:

Serializing
- if human readable serializer, always hex
- if non-human readable, always write bytes

Deserializing
- Accept bytearray/buf
- Accept Hex string with & w/o 0x prefix
- Accept sequence containing bytes

Does that seem reasonable?

DaniPopes: yeah makes sense

crates/primitives/src/bits/serde.rs
```diff
@@ -10,13 +14,83 @@ impl<const N: usize> Serialize for FixedBytes<N> {
...
+                }
+
+                Ok(FixedBytes(bytes))
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                let v = v.strip_prefix("0x").unwrap_or(v);
```
DaniPopes, line 67: already checked in hex

crates/primitives/src/bits/serde.rs
```diff
@@ -10,13 +14,83 @@ impl<const N: usize> Serialize for FixedBytes<N> {
...
+                    .map_err(|_| {
+                        de::Error::invalid_length(
+                            v.len(),
+                            &format!("exactly {} bytes, as {} hex chars", N, N * 2).as_str(),
+                        )
+                    })
+                    .map(FixedBytes)
+            }
+        }
+        deserializer.deserialize_any(FixedVisitor::<N>)
```
DaniPopes, line 82: so we want to check for Human readable?

prestwich: i think that we should on Serialize. but on deserialize, because the only human-readable we expect is also self-describing (json), I think it's unnecessary. the serde_json deserializer will not treat a string as bytes unless explicitly instructed to via `deserialize_bytes`. That implies that checking for human-readable is unnecessary

prestwich: code ref: https://github.com/serde-rs/json/blob/master/src/de.rs#L1416-L1423

prestwich:
Gave the same treatment to `Bytes` as well

crates/primitives/src/bytes/serde.rs
```diff
@@ -1,23 +1,84 @@
...
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                hex::decode(v)
+                    .map_err(|_| {
+                        de::Error::invalid_value(de::Unexpected::Str(v), &"a valid hex string")
+                    })
+                    .map(From::from)
```
DaniPopes, line 66:
```suggestion
                v.parse()
                    .map_err(|_| {
                        de::Error::invalid_value(de::Unexpected::Str(v), &"a valid hex string")
                    })
```

crates/primitives/src/bytes/serde.rs
```diff
@@ -1,23 +1,84 @@
...
+                E: serde::de::Error,
+            {
+                Ok(Bytes::from(v))
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mut bytes = Vec::new();
```
DaniPopes, line 49: [`seq.size_hint() -> Option<usize>`](https://docs.rs/serde/latest/serde/de/trait.SeqAccess.html#method.size_hint)

DaniPopes (changes requested):
I'll implement these rq

DaniPopes (approved):

Issue #201, opened 2023-07-20
[Bug] sol! macro doesn't alphabetize nested struct postfixes in SolStruct::eip712_encode_type()


nhtyy:
### Component

sol-type

### What version of Alloy are you on?

├── alloy-sol-types v0.2.0 │   ├── alloy-primitives v0.2.0 │   ├── alloy-sol-macro v0.2.0 (proc-macro)

### Operating System

macOS (Apple Silicon)

### Describe the bug

```solidity
use alloy_sol_types::sol;

sol! {
    /// @dev generic order information
    ///  should be included as the first field in any concrete order types
    struct OrderInfo {
        // The address of the reactor that this order is targeting
        // Note that this must be included in every order so the swapper
        // signature commits to the specific reactor that they trust to fill their order properly
        address reactor;
        // The address of the user which created the order
        // Note that this must be included so that order hashes are unique by swapper
        address swapper;
        // The nonce of the order, allowing for signature replay protection and cancellation
        uint256 nonce;
        // The timestamp after which this order is no longer valid
        uint256 deadline;
        // Custom validation contract
        address additionalValidationContract;
        // Encoded validation params for additionalValidationContract
        bytes additionalValidationData;
    }

    /// @dev An amount of output tokens that decreases linearly over time
    struct DutchOutput {
        // The ERC20 token address (or native ETH address)
        address token;
        // The amount of tokens at the start of the time period
        uint256 startAmount;
        // The amount of tokens at the end of the time period
        uint256 endAmount;
        // The address who must receive the tokens to satisfy the order
        address recipient;
    }

    /// @dev An amount of input tokens that increases linearly over time
    struct DutchInput {
        // The ERC20 token address
        address token;
        // The amount of tokens at the start of the time period
        uint256 startAmount;
        // The amount of tokens at the end of the time period
        uint256 endAmount;
    }

    struct ExclusiveDutchOrder {
        // generic order information
        OrderInfo info;
        // The time at which the DutchOutputs start decaying
        uint256 decayStartTime;
        // The time at which price becomes static
        uint256 decayEndTime;
        // The address who has exclusive rights to the order until decayStartTime
        address exclusiveFiller;
        // The amount in bps that a non-exclusive filler needs to improve the outputs by to be able to fill the order
        uint256 exclusivityOverrideBps;
        // The tokens that the swapper will provide when settling the order
        DutchInput input;
        // The tokens that must be received to satisfy the order
        DutchOutput[] outputs;
    }
}

[...]

PR #203, opened 2023-07-20
fix: correct encodeType expansion for nested structs

prestwich:
## Motivation

#201 

## Solution

- introduce `eip712_root_type` and `eip712_component_types` to get typestrings on the fly
- sort, dedup, and append typestring in default impl of `eip712_encode_type`

This is a bit clunky. I don't love it

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro/src/expand/struct.rs
+++ b/crates/sol-macro/src/expand/struct.rs
@@ -40,22 +40,7 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, s: &ItemStruct) -> Result<TokenStream> {
         .map(|f| (expand_type(&f.ty), f.name.as_ref().unwrap()))
         .unzip();
 
-    let encoded_type = fields.eip712_signature(name.as_string());
-    let encode_type_impl = if fields.iter().any(|f| f.ty.is_custom()) {
-        quote! {
-            {
-                let mut encoded = String::from(#encoded_type);
-                #(
-                    if let Some(s) = <#field_types as ::alloy_sol_types::SolType>::eip712_encode_type() {
-                        encoded.push_str(&s);
-                    }
-                )*
-                encoded
-            }
-        }
-    } else {
-        quote!(#encoded_type)
-    };
+    let eip712_encode_type_fns: TokenStream = expand_encode_type_fns(fields, name);
 
     let tokenize_impl = expand_tokenize_func(fields.iter());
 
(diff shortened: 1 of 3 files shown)


crates/sol-types/src/types/struct.rs
```diff
@@ -63,9 +63,29 @@ pub trait SolStruct: 'static {
...
+    fn eip712_encode_type() -> Cow<'static, str> {
+        let mut components = Self::eip712_components();
+        components.sort();
+        components.dedup();
+
+        Cow::Owned(
+            std::iter::once(Self::eip712_root_type())
+                .chain(components.into_iter())
+                .collect::<Vec<_>>()
+                .join(""),
```
DaniPopes, line 86: can we collect into a string directly?

crates/sol-types/src/types/struct.rs
```diff
@@ -63,9 +63,29 @@ pub trait SolStruct: 'static {
...
+    /// `encodeType` string.
+    fn eip712_root_type() -> Cow<'static, str>;
+
     /// EIP-712 `encodeType`
     /// <https://eips.ethereum.org/EIPS/eip-712#definition-of-encodetype>
-    fn eip712_encode_type() -> Cow<'static, str>;
+    fn eip712_encode_type() -> Cow<'static, str> {
+        let mut components = Self::eip712_components();
+        components.sort();
+        components.dedup();
```
DaniPopes, line 80: empty check to early return?

crates/sol-types/src/types/struct.rs
```diff
@@ -63,9 +63,29 @@ pub trait SolStruct: 'static {
         self.tokenize().total_words() * Word::len_bytes()
     }
 
+    /// Returns component EIP-712 types. These types are used to construct
+    /// the `encodeType` string. These are the types of the struct's fields,
+    /// and should not include the root type.
+    fn eip712_components() -> Vec<Cow<'static, str>>;
```
DaniPopes, line 69: should this even be exposed? are there any other use cases

crates/sol-macro/src/expand/struct.rs
```diff
@@ -151,3 +134,57 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, s: &ItemStruct) -> Result<TokenStream> {
...
+
+    if bits.is_empty() {
+        quote! {vec![]}
+    } else {
+        quote! {
+            let mut components = vec![];
+            #(#bits)*
+            components
+        }
+    }
```
DaniPopes, line 189: use fully qualified path for vec macro and <#ty as ..>

DaniPopes (changes requested):

crates/sol-types/src/types/struct.rs
prestwich: unless we go through some extra trait wrangling, we can't both
- use it in the default impl of `eip712_encode_type`
- not expose it

Is there a specific setup you'd prefer?

DaniPopes (approved):

PR #223, opened 2023-08-02
fix bincode serialization

### LINKED ISSUE #221: [Bug] `bincode` deserialization fails for fixed-size byte arrays
### Component

primitives

alloy-primitives v0.3.1

### Operating System

None

### Describe the bug

Trying to use `bincode` to deserialize `FixedBytes<N>` fails because `deserializer.deserialize_any` is not supported.

Code snippet to reproduce this bug:
```rust
let hash = B256::ZERO;
bincode::deserialize::<B256>(&bincode::serialize(&hash).unwrap()).unwrap();
```

The `serde` documentation also discourages the use of `deserialize_any` for this very reason:
> When implementing Deserialize, you should avoid relying on Deserializer::deserialize_any unless you need to be told by the Deserializer what type is in the input. Know that relying on Deserializer::deserialize_any means your data type will be able to deserialize from self-describing formats only, ruling out Postcard and many others.

However, as I understand #202, JSON deserialization support for either a byte array, byte sequence, or string was explicitly required. Here I see a possible compromise to also check `is_human_readable()` in the deserialization and keep the current parsing only for human readable, e.g. JSON, and use `deserializer.deserialize_bytes` otherwise, which would also allow `bincode` and similar formats.

If this approach is acceptable, I would be happy to provide a PR.

Wollac:
## Motivation

Fixes `bincode` deserialization of `FixedBytes` and `Bytes`.

## Solution

Distinguish between `is_human_readable` and not also in the deserialization. If it is not human readable (e.g. `bincode`) allow only bytes instead of also strings.
This fixes #221 .

Added tests for the `bincode` roundtrip that previously failed.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/serde.rs
+++ b/crates/primitives/src/bits/serde.rs
@@ -18,58 +18,82 @@ impl<const N: usize> Serialize for FixedBytes<N> {
 
 impl<'de, const N: usize> Deserialize<'de> for FixedBytes<N> {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        struct FixedVisitor<const N: usize>;
+        if deserializer.is_human_readable() {
+            struct FixedVisitor<const N: usize>;
 
-        impl<'de, const N: usize> Visitor<'de> for FixedVisitor<N> {
-            type Value = FixedBytes<N>;
+            impl<'de, const N: usize> Visitor<'de> for FixedVisitor<N> {
+                type Value = FixedBytes<N>;
 
-            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-                write!(
-                    formatter,
+                fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                    write!(
+                        formatter,
                     "{} bytes, represented as a hex string of length {}, an array of u8, or raw bytes",
-                    N,
-                    N * 2
-                )
-            }
-
-            fn visit_bytes<E: de::Error>(self, v: &[u8]) -> Result<Self::Value, E> {
-                <[u8; N]>::try_from(v)
-                    .map(FixedBytes)
-                    .map_err(de::Error::custom)
-            }
-
-            fn visit_seq<A: de::SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
-                let mut bytes = [0u8; N];
-
-                bytes.iter_mut().enumerate().try_for_each(|(i, b)| {
-                    *b = seq.next_element()?.ok_or_else(|| {
-                        de::Error::invalid_length(i, &format!("exactly {} bytes", N).as_str())
-                    })?;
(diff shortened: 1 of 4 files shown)


crates/primitives/src/bits/serde.rs
```diff
@@ -18,58 +18,82 @@ impl<const N: usize> Serialize for FixedBytes<N> {
 
 impl<'de, const N: usize> Deserialize<'de> for FixedBytes<N> {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        struct FixedVisitor<const N: usize>;
+        if deserializer.is_human_readable() {
```
prestwich, line 21: couldn't this be simplified to avoid the second visitor definition?

```
if deserializer.is_human_readable() {
    deserializer.deserialize_any(FixedVisitor::<N>)
} else {
    deserializer.deserialize_bytes(FixedVisitor::<N>)
```

Wollac: Very good point. I'll doublecheck to see if anything weird happens with those then unused methods, but I don't suspect it.

Wollac: changed

crates/primitives/src/bytes/serde.rs
```diff
@@ -53,13 +53,18 @@ impl<'de> serde::Deserialize<'de> for Bytes {
...
+        } else {
+            deserializer.deserialize_byte_buf(BytesVisitor)
+        }
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bincode as _;
```
DaniPopes, line 67: Nit: I don't think this does anything

```suggestion
```

DaniPopes (changes requested):
LGTM after nit and rebase to fix clippy

crates/primitives/src/bytes/serde.rs
Wollac: Without those I got:
```
warning: external crate `bincode` unused in `alloy_primitives`: remove the dependency or add `use bincode as _;`
```
When building the tests without the `serde` feature.
I don't know how to prevent this warning. Is there a better way?

DaniPopes: Ah I see, it's fine then. The problem is that it gets included in the test binaries but not in the library, so it flags it as unused in the library.

Wollac: So we decided to have the `use` in? In both places?
What I don't really understand is why the same is not needed for `serde_json`. Seems like it's only used in the serde tests as well...

DaniPopes: It is: https://github.com/alloy-rs/core/blob/4f7a72c476de6b008be275daaa577ef9e89156e5/crates/primitives/src/lib.rs#L20-L24

DaniPopes (approved):
Thanks!

PR #226, opened 2023-08-05
fix(sol-macro): snake_case'd function names

DaniPopes:
## Motivation

`heck` doesn't treat numbers as new words, and discards leading underscores, thus:

```solidity
contract C {
  function f();
  function _f();
}
```

failed to compile because the generated `is_`... functions would have the same name for both variants.

## Solution

Create a helper function for snakifying idents, taken from [`strum`](https://github.com/Peternator7/strum/blob/fcb9841a744543b993eb1645db35dbda541e9a30/strum_macros/src/macros/enum_is.rs#L48C1-L61C2)

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


prestwich (approved):

prestwich:
just cliipy

Issue #210, opened 2023-07-26
[Feature] Simple API to resolve `json-abi` item types


DaniPopes:
### Component

dyn-abi, json-abi

### Describe the feature you would like

Example: in `ethabi` you can just call `Function::encode_input` to ABI-encode data with the function's selector and inputs. cc @prestwich

davebryson:
I'm experimenting with this now to get a better understanding of alloy's APIs. What's the intended replacement for`ethabi Token`  in this context ... `SolType`?  Thanks!

prestwich:
DynSolType is roughly equivalent to ParamKind

The way i would to this is extending the Resolve trait in dyn-abi to have a method that both resolves + encodes

davebryson:
Here's one approach as a method on [Function](https://github.com/alloy-rs/core/blob/main/crates/json-abi/src/item.rs#L376)

```rust
pub fn encode_input(&self, args: DynAbiValue) -> Result<Vec<u8>> {
   let input_types = self
            .inputs
            .iter()
            .map(|i| i.resolve().expect("valid type"))
            .collect::<Vec<DynSolType>>();
   
    if !input_types.matches(&args) {
            Err(...)
    }
      
    let encoded_args = args.encode_params();
    // prepend selector... etc...
}
```
I'm actually working on a pull request for this.  However,  this approach requires `json_abi` to depend on `dyn_abi` which already depends on `json_abi`.... which depends on ... resulting in a cyclic deps problem.   Any thoughts on how to get around this?

prestwich:
yeah, the circular dep is why we need to put this logic in the `Resolve` trait in `dyn-abi`. json-abi is strictly a dat format, and should not rely on the type system representation in `dyn-abi`

davebryson:
@prestwich Ok, I think I understand what you're after now.

davebryson:
@DaniPopes  Following on to this, is there interest in taking a similar approach to provide support for something like `event.parse_log()` to parse the log into DynSolValues?

prestwich:
@DaniPopes should we come up with a struct that represents Events? 

```
struct DynEventType {
    topics: [Option<DynSolType>; 4],
    body: DynSolType
}

struct DynEvent {
    topics: [Option<DynSolValue>; 4],
    body: DynSolValue
}
```

davebryson:
@DaniPopes @prestwich  I'm happy to work on this once you decide on the direction.  I've already roughed out an initial solution

PR #227, opened 2023-08-07
feat: Add support for function input/output encoding/decoding

davebryson:
## Motivation
Added support for Function input encoding and output decoding.  This is in reference to issue #210 

## Solution
Added trait `FunctionExt` that defines 2 methods: `encode_input(...)` and `decode_output(...)` and an implementation of the trait for `json_abi::Function`.  Currently this resides in `resolve.rs`.

## PR Checklist
- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/dyn-abi/Cargo.toml
+++ b/crates/dyn-abi/Cargo.toml
@@ -41,11 +41,25 @@ hex-literal.workspace = true
 criterion.workspace = true
 ethabi = "18"
 rand = "0.8"
+serde_json = { workspace = true }
 
 [features]
 default = ["std"]
-std = ["alloy-json-abi/std", "alloy-sol-types/std", "alloy-sol-type-parser/std", "alloy-primitives/std", "hex/std", "serde?/std", "serde_json?/std"]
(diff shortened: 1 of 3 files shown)


prestwich:
discussing this approach internally and will loop back

davebryson:
As I dig in a bit more, to grok all of `core`, I'm wondering if `SolType` is intended as a (better) replacement for  ethabi `Token`.  If that's the case, should `encode_input()` take a `SolType` as arguments vs `DynSolValue`?

DaniPopes:
`ethabi::ParamType` -> `DynSolType`
`ethabi::Token` -> `DynSolValue`

davebryson:
> `ethabi::ParamType` -> `DynSolType` `ethabi::Token` -> `DynSolValue`

Ok. I'm on the right track then. However, at a higher level, I can still envision the need for something like `ethers::Tokenizable` to go `to` and `from` `DynSolValue`.

prestwich:
implementing `Into<DynSolValue>` seems to be sufficient? 🤔

DaniPopes (approved):
Thanks, will touch up this PR in a follow-up.

PR #228, opened 2023-08-13
feat: implement abi2sol

DaniPopes:
## Motivation

Rewrite code that generates Solidity tokens from a `JsonAbi` in `sol-macro` to use strings in `json-abi`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [no] Breaking changes

### DIFF
--- a/crates/json-abi/src/abi.rs
+++ b/crates/json-abi/src/abi.rs
@@ -33,6 +33,34 @@ pub struct JsonAbi {
 }
 
 impl JsonAbi {
+    /// Creates an empty ABI object.
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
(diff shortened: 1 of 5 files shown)


prestwich (commented):
it seems like we want to unify this to_sol with syn-solidity somehow, no? rather than assembling strings like this, we should consider converting abi json to syn-solidity types and then having a pipeline for converting syn-solidity into sol files?

prestwich:
my point is that "converting type specifications to solidity strings" is probably a common action that we should capture in a single location. should we capture it here, or elsewhere?

DaniPopes:
@prestwich thinking too far ahead :) This is fine for now.

prestwich (approved):

PR #231, opened 2023-08-15
perf: optimize some stuff

DaniPopes:
- `try_for_each` generally generates more llvm-ir and tends to be less optimized than just loop with `?`
- rest of changes are commented below individually

### DIFF
--- a/crates/dyn-abi/src/token.rs
+++ b/crates/dyn-abi/src/token.rs
@@ -7,7 +7,7 @@ use alloy_sol_types::token::{PackedSeqToken, TokenType, WordToken};
 // NOTE: do not derive `Hash` for this type. The derived version is not
 // compatible with the current `PartialEq` implementation. If manually
 // implementing `Hash`, ignore the `template` prop in the `DynSeq` variant
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug)]
 pub enum DynToken<'a> {
     /// A single word.
     Word(Word),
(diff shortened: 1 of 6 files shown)


crates/sol-types/src/coder/encoder.rs
```diff
@@ -153,6 +128,31 @@ impl Encoder {
...
+            ptr::copy_nonoverlapping(bytes.as_ptr(), dst, cnt);
+
+            // set remaining bytes to zero if necessary
+            let rem = cnt % 32;
+            if rem != 0 {
+                let pad = 32 - rem;
+                ptr::write_bytes(dst.add(cnt), 0, pad);
+            }
+        }
+    }
```
DaniPopes, line 155: `append_bytes` copied 32 bytes at a time, when it can just copy everything at once

crates/sol-types/src/types/data_type.rs
```diff
@@ -260,7 +260,7 @@ impl SolType for Function {
 /// Bytes - `bytes`
 pub struct Bytes;
 
-impl<T: AsRef<[u8]>> Encodable<Bytes> for T {
+impl<T: ?Sized + AsRef<[u8]>> Encodable<Bytes> for T {
```
DaniPopes, line 263: allows:

```rust
let x: &[u8] = &[];
Encodable::<Bytes>::to_tokens(x)
```

crates/dyn-abi/src/token.rs
```diff
@@ -154,23 +154,31 @@ impl<'a> DynToken<'a> {
             Self::DynSeq { contents, template } => {
                 let mut child = dec.take_indirection()?;
                 let size = child.take_u32()? as usize;
+                if size == 0 {
+                    // should already be empty from `empty_dyn_token`
+                    debug_assert!(contents.is_empty());
+                    return Ok(())
+                }
+
```
DaniPopes, line 162: early return for empty sequence

crates/dyn-abi/src/token.rs
```diff
@@ -154,23 +154,31 @@ impl<'a> DynToken<'a> {
...
-                    .take()
-                    .expect("No template. This is a bug, please report it.");
-                new_tokens.resize(size, *t);
+                let t = template.take().expect("no template for dynamic sequence");
+                let mut new_tokens = if size == 1 {
+                    // re-use the box allocation
+                    unsafe { Vec::from_raw_parts(Box::into_raw(t), 1, 1) }
+                } else {
+                    vec![*t; size]
+                };
```
DaniPopes, line 177: re use box allocation

prestwich (approved):

PR #233, opened 2023-08-20
fix(json-abi): properly handle Param `type` field

DaniPopes:
## Motivation

`type` is either:
- "tuple" + [] array brackets if `components` is not empty
- a canonical Solidity type if `components` is empty

Check this during deserialization and signature formatting.

Added benchmark for signature formatting:

```
signature/large-function/alloy                                                                            
                        time:   [217.91 ns 218.47 ns 219.13 ns]
signature/large-function/ethabi                                                                             
                        time:   [2.6704 µs 2.6737 µs 2.6768 µs]
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/json-abi/src/param.rs
```diff
@@ -184,6 +180,12 @@ impl Param {
             s.push_str(&self.ty)
         } else {
             crate::utils::signature_raw("", &self.components, s);
+            // checked during deserialization, but might be invalid from a user
+            debug_assert!(
+                self.ty.starts_with("tuple"),
+                "expected tuple param type to start with \"tuple\"",
+            );
+            s.push_str(unsafe { self.ty.get_unchecked(5..) });
```
DaniPopes, line 188: Should we allow this with `if let Some = strip_prefix` ?

crates/json-abi/src/param.rs
```diff
@@ -184,6 +180,12 @@ impl Param {
             s.push_str(&self.ty)
         } else {
             crate::utils::signature_raw("", &self.components, s);
+            // checked during deserialization, but might be invalid from a user
```
prestwich, line 183: could we check this on type-instantiation so it's impossible for a user to construct an invalid one?

DaniPopes: How? All struct fields are public

prestwich: if public fields allow the user to create invalid instances, then we should at least consider making them non-public

prestwich (approved):

PR #250, opened 2023-08-28
feat(sol-macro): add opt-in attributes for extra methods and derives

DaniPopes:
## Motivation

Helps with compilation speed, most of the time these are not used. In the future it's possible to further expand these attributes to declare a list of item names to include, instead of adding it to either none or all items.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


prestwich (approved):
dope

PR #260, opened 2023-09-04
feat(syn-solidity): improve variable getters generation

### LINKED ISSUE #253: [Feature] Generate getters for public mapping variables
### Component

sol! macro

Currently this code:

https://github.com/alloy-rs/core/blob/5267aae3fe588e6c023032a622edbe7d86693341/crates/syn-solidity/src/item/function.rs#L101-L103

Only handles the basic case of `uint256 public x = ...;` -> `function x() returns (uint256)`

This can be expanded to support:
`mapping(type1 key1 => mapping(type2 key2 => ... valueType value)) public map = ...;`
-> `function map( $($type $key),* ) returns (valueType value)`

### Additional context

_No response_

DaniPopes:
## Motivation

Closes #253

## Solution

Recurse into mappings in value positions, adding keys to the arguments list, and setting the last non-mapping value as the return value.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/syn-solidity/src/attribute/function.rs
+++ b/crates/syn-solidity/src/attribute/function.rs
@@ -106,10 +106,6 @@ impl FunctionAttributes {
         self.0.iter().any(FunctionAttribute::is_virtual)
     }
 
-    pub fn has_immutable(&self) -> bool {
-        self.0.iter().any(FunctionAttribute::is_immutable)
-    }
-
     pub fn has_override(&self, path: Option<&SolPath>) -> bool {
         self.0.iter().any(|attr| attr.is_override(path))
     }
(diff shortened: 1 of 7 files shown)


crates/syn-solidity/src/attribute/function.rs
```diff
@@ -155,7 +148,6 @@ impl fmt::Debug for FunctionAttribute {
             Self::Visibility(visibility) => f.debug_tuple("Visibility").field(visibility).finish(),
             Self::Mutability(mutability) => f.debug_tuple("Mutability").field(mutability).finish(),
             Self::Virtual(_) => f.write_str("Virtual"),
-            Self::Immutable(_) => f.write_str("immutable"),
```
prestwich, line 158: drve-by nit: uncaptialize `"Virtual"` on the line above

crates/syn-solidity/src/attribute/function.rs
```diff
@@ -207,11 +197,15 @@ impl Parse for FunctionAttribute {
 }
 
 impl From<VariableAttribute> for FunctionAttribute {
+    /// Converts a variable attribute to its corresponding function attribute.
+    ///
+    /// - `constant` -> `pure`
+    /// - `immutable` -> `view`
```
prestwich, line 203: both `constant` and `immutable` correspond to code regions and `codecopy` commands, no? shoudn't they be the same?

crates/syn-solidity/src/attribute/function.rs
DaniPopes: It's debug fmt

crates/syn-solidity/src/attribute/function.rs
DaniPopes: Doesn't really matter, just needed a way to differentiate the two

prestwich (approved):

PR #263, opened 2023-09-10
feat(sol-macro): expand fields with attrs

0xgregthedev:
## Motivation
When using a derive macro inside of the sol! macro, field attributes are not propagated to the expanded item.

## Solution
Added 'attrs' parameter to 'expand_field'
Passed 'attrs' arg to 'expand_field' in 'expand_fields'
Moved 'pub' visibility specification to 'expand_field' 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
Thanks!

PR #268, opened 2023-09-13
Support yul ast 

### LINKED ISSUE #237: [Feature] Support `yul` AST
### Component

syn-solidity

`yul` items are the only missing items from `syn-solidity`.

All nodes must implement:
- `std::fmt::Debug`
- `syn::parse::Parse`
- `crate::Spanned`

and should have a short 1 liner doc comment, optionally pointing to the [Solidity grammar reference](https://docs.soliditylang.org/en/latest/grammar.html) where possible.

### Additional context

See #199 for reference of how to complete the AST.

mouseless-eth:
## Motivation

Closes #237

## Solution

Added Yul types as specified by the [Solidity Language Grammar docs](https://docs.soliditylang.org/en/latest/grammar.html#language-grammar).

## Notice

Parsing correctness has not been verified (not sure how to approach this).

Enforcement of syntax rules has not been accounted for, e.g., [yul decimal numbers cannot start with leading zeros](https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityLexer.YulDecimalNumber).

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/syn-solidity/src/yul/expr/fn_call.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl fmt::Debug for YulFnCall {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("YulFnCall")
+            .field("function_type", &self.function_type)
+            .field("paren_token", &self.paren_token)
+            .field("arguments", &self.arguments)
+            .finish()
+    }
+}
```
DaniPopes, line 56: The purpose of manually implementing `Debug` was to skip the token fields since they don't add anything useful to the output

crates/syn-solidity/src/lib.rs
```diff
@@ -84,7 +84,7 @@ pub mod visit_mut;
 pub use visit_mut::VisitMut;
 
 mod yul;
-pub use yul::YulBlock;
+pub use yul::*;
```
DaniPopes, line 87: I know it's a chore, but can you list all items instead of glob? On VS Code I use `loop { <Comma>; Ctrl + I; Enter }` to write all the items that can be exported until you reach `self`

crates/syn-solidity/src/yul/ident/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+mod path;
+pub use path::YulPath;
+
+/// A Yul identifier.
+///
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityLexer.YulIdentifier>
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct YulIdent(pub Ident);
```
DaniPopes, line 21: Are these different in any way to the regular Solidity ones? The grammar def is the same, but I'm wondering if the keyword handling is different that it merits a separate type

crates/syn-solidity/src/yul/ident/path.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulPath>
+#[derive(Clone)]
+pub enum YulPath {
+    /// Dotless path, references a declared varaible inside the assembly block.
+    SimplePath(YulIdent),
+
+    /// Dotted path, references a declared variable outside the assembly block.
+    DottedPath(Punctuated<YulIdent, Token![.]>),
+}
```
DaniPopes, line 23: This should just be a wrapper struct for `Punctuated<YulIdent, Token![.]>`, the enum seems unnecessary

crates/syn-solidity/src/yul/lit/hex_num.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// Represents the `0x` prefix/token
+#[derive(Clone)]
+pub struct ZeroExPrefix {
+    zero_token: LitInt,
+    x_token: Punct,
+}
+
+impl Parse for ZeroExPrefix {
```
DaniPopes, line 57: This parses multiple tokens: "0 x"
Just use the literals in `crate::lit`

crates/syn-solidity/src/yul/lit/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+
+mod hex_num;
+pub use hex_num::YulHexNum;
+
+/// Yul literals e.g. 0x123, 42 or "abc".
+///
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulLiteral>
+#[derive(Clone)]
+pub enum YulLit {
```
DaniPopes, line 18: These are similar enough to Rust and Solidity that you can just use `crate::Lit`

crates/syn-solidity/src/yul/stmt/assignment.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Assign a single variable.
+    Single(YulSingleAssign),
+
+    /// Assign many variables, only assignable via function call.
+    Multi(YulMultiAssign),
+}
+
+impl Parse for YulVarAssign {
+    fn parse(input: ParseStream<'_>) -> Result<Self> {
+        if input.peek2(Token![,]) {
```
DaniPopes, line 28: This is slightly wrong since you can assign to paths and so the comma is not always the second token

crates/syn-solidity/src/yul/stmt/assignment.rs
```diff
@@ -0,0 +1,144 @@
...
+    punctuated::Punctuated,
+    Result, Token,
+};
+
+/// Assignaration of one or more Yul variables with optional initial value.
+/// If multiple variables are declared, only a function call is a valid initial
+/// value.
+///
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulVariableAssignaration>
```
DaniPopes, line 16: correct link is this <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulAssignment>

crates/syn-solidity/src/yul/stmt/assignment.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Clone)]
+pub enum YulVarAssign {
```
DaniPopes, line 18: I'm thinking this should be a struct instead, with LHS a punctuated of paths

crates/syn-solidity/src/yul/stmt/switch.rs
```diff
@@ -0,0 +1,156 @@
...
+/// ```
+///
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulSwitchStatement>
+#[derive(Clone)]
+pub struct YulSwitch {
+    switch_token: kw::switch,
+    selector: YulExpr,
+    branches: Vec<YulCaseBranch>,
+    default_case: Option<YulSwitchDefault>,
```
DaniPopes, line 26: missing pub

crates/syn-solidity/src/yul/stmt/switch.rs
```diff
@@ -0,0 +1,156 @@
...
+            .finish()
+    }
+}
+
+/// Represents a non-default case of a Yul switch statement.
+#[derive(Clone)]
+pub struct YulCaseBranch {
+    case_token: kw::case,
+    constant: YulLit,
+    body: YulBlock,
```
DaniPopes, line 85: missing pub

crates/syn-solidity/src/yul/stmt/switch.rs
```diff
@@ -0,0 +1,156 @@
...
+            .field("body", &self.body)
+            .finish()
+    }
+}
+
+/// Represents the default case of a Yul switch statement.
+#[derive(Clone)]
+pub struct YulSwitchDefault {
+    default_token: kw::default,
+    body: YulBlock,
```
DaniPopes, line 125: missing pub

crates/syn-solidity/src/yul/stmt/switch.rs
```diff
@@ -0,0 +1,156 @@
...
+                branches
+            },
+            default_case: {
+                if input.peek(kw::default) {
+                    Some(input.parse()?)
+                } else {
+                    None
+                }
+            },
+        })
```
DaniPopes, line 48: from the spec looks like we need a default case if no other cases are provided, but here we may skip it. I think this is fine

crates/syn-solidity/src/yul/stmt/var_decl.rs
```diff
@@ -0,0 +1,162 @@
...
+    punctuated::Punctuated,
+    Result, Token,
+};
+
+/// Declaration of one or more Yul variables with optional initial value.
+///
+/// Solidity Reference:
+/// <https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityParser.yulVariableDeclaration>
+#[derive(Clone)]
+pub enum YulVarDecl {
```
DaniPopes, line 16: Same as assignment

crates/syn-solidity/src/yul/type/function.rs
```diff
@@ -0,0 +1,118 @@
...
+            .field("ident", &self.ident)
+            .field("paren_token", &self.paren_token)
+            .field("arguments", &self.arguments)
+            .field("returns", &self.returns)
+            .field("body", &self.body)
+            .finish()
+    }
+}
+
+/// The return attribute of a Yul function defenition.
```
DaniPopes, line 70:
```suggestion
/// The return attribute of a Yul function definition.
```

DaniPopes (changes requested):
First of all thank you so much for taking this on!

We have to get this right the first time to avoid major breaking changes in the future, I hope you don't mind the amount of comments :smile_cat:

crates/syn-solidity/src/yul/ident/mod.rs
mouseless-eth: Solidity and Yul handle keywords differently because they each have their own unique sets of keywords and reserved identifiers. 

[solidity reserved  identifiers:](https://github.com/ethereum/solidity/blob/020b59680ebce25367ece2c1388584a8c3827d85/liblangutil/Token.h#L235) `after`, `alias`, `apply`, `auto`, `byte`, ...

[yul reserved identifiers:](https://github.com/ethereum/solidity/blob/020b59680ebce25367ece2c1388584a8c3827d85/libyul/backends/evm/EVMDialect.cpp#L114) `basefee`, `prevrandao`, `linkersymbol`, `datasize`, `dataoffset`, `datacopy`, ...

[solidity keywords](https://github.com/ethereum/solidity/blob/020b59680ebce25367ece2c1388584a8c3827d85/liblangutil/Token.h#L142)

[yul keywords](https://github.com/ethereum/solidity/blob/020b59680ebce25367ece2c1388584a8c3827d85/liblangutil/Token.h#L321) 

Are keyword checks planned for implementation in the future? if not, then can reuse the Solidity ident type for Yul.

DaniPopes: Thanks for the links, I'm still thinking about it. This wrapper type is fine for now

mouseless-eth:
I believe all the changes have now been addressed.

crates/syn-solidity/src/yul/ident/mod.rs
mouseless-eth: Leaving it as is for now. The code duplication feels a little bit off, but curious to hear what you come up with

DaniPopes (approved):
Thanks!

PR #271, opened 2023-09-15
feat: add logs, add log dynamic decoding

### LINKED ISSUE #210: [Feature] Simple API to resolve `json-abi` item types

DaniPopes:
## Motivation

Closes #210

Common `Log` structure

## Solution

- Add `Log` structure
- Dynamic coding support and add `EventExt` trait

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/primitives/src/log.rs
```diff
@@ -0,0 +1,34 @@
...
+/// An Ethereum event log object.
+#[derive(Clone, Default, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "arbitrary",
+    derive(derive_arbitrary::Arbitrary, proptest_derive::Arbitrary)
+)]
+pub struct Log {
+    /// The indexed topic list.
+    pub topics: Vec<B256>,
```
prestwich, line 13: `pub` allows for construction of invalid logs with 5+ topics

crates/sol-types/src/types/event/mod.rs
```diff
@@ -151,4 +151,9 @@ pub trait SolEvent: Sized {
         let body = Self::decode_data(data, validate)?;
         Ok(Self::new(topics, body))
     }
+
```
prestwich, line 154: hot take. We should consider `#[doc(hide)]` all the trait methods that do _not_ operate on `&Log` or output `Log`, as users should generally not use them directly

crates/dyn-abi/src/ext/event.rs
```diff
@@ -0,0 +1,301 @@
...
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for alloy_json_abi::Event {}
+}
+use sealed::Sealed;
+
+/// Provides event encoding and decoding for the [`Event`] type.
+///
+/// This trait is sealed and cannot be implemented for types outside of this
+/// crate. It is implemented only for [`Event`].
```
prestwich, line 16: Should this be a `Resolve` like interface that outputs a `DynEvent` and then uses that to (en/de)code?

prestwich (changes requested):
1 change, a couple interface ideas/questions

crates/sol-types/src/types/event/mod.rs
DaniPopes: They should because `Log` can't borrow data while the other can

crates/dyn-abi/src/ext/event.rs
DaniPopes: The only thing it would do is resolve the types, I don't think so

crates/primitives/src/log.rs
DaniPopes: Skill issue

prestwich: types exist to solve skill issues :P

gakonst: i would agree we can add getters instead

crates/dyn-abi/src/ext/event.rs
prestwich: i guess this is a question of how we conceptualize dyn-abi. if it has the same scope as sol-types then we should have dyn equivalents to everything in sol-types

the original idea was to have a runtime representation that was format-independent, and then be able to convert JSON or solidity lines, or type strings to the runtime representation. we're now kinda mixing that, where _some_ things are resolved to the runtime repr, and some aren't. which leads to a somewhat weird API

crates/dyn-abi/src/ext/abi.rs
```diff
@@ -92,8 +92,8 @@ impl JsonAbiExt for Constructor {
     }
 
     #[inline]
-    fn decode_input(&self, data: &[u8]) -> Result<Vec<DynSolValue>> {
-        decode(data, &self.inputs)
+    fn decode_input(&self, data: &[u8], validate: bool) -> Result<Vec<DynSolValue>> {
```
gakonst, line 95: What's our general stance on functions which have `validate: bool`? When is it OK to not validate?

crates/dyn-abi/src/ext/event.rs
prestwich: or for a concrete thing, if we do this as an ext trait then we duplicate code across these two impls
```
impl EventExt for Event {}
impl EventExt for syn_solidity::ItemEvent {}
```

I would prefer something like

```
trait ResolveEvent {
    fn resolve(&self) -> DynEvent
}

impl ResolveEvent for JsonEvent {}
impl ResolveEvent for syn_soldity::ItemEvent {}

```

crates/dyn-abi/src/ext/event.rs
```diff
@@ -0,0 +1,301 @@
...
+                "
+			0000000000000000000000000000000000000000000000000000000000012345
+			0000000000000000000000000000000000000000000000000000000000054321
+			"
+            ),
+        );
+
+        wrong_event.decode_log_object(&log, false).unwrap();
+        // TODO: How do we verify here?
+        // wrong_event.decode_log_object(&log, true).unwrap_err();
```
gakonst, line 297: what do we want to check?

gakonst (commented):
supportive w/o many opinions beyond maybe providing a `LogMeta`-style type which may include more information? e.g. thinking Revm addresses, thinking how in ethers-providers we returned logs w/ txhashes etc. in them

crates/primitives/src/log.rs
DaniPopes: They might be invalid in EVM, but we can use invalid states for testing and maybe other usecases in the futures. I don't think we should lock ourselves into using arrays since it doesn't really improve anything, might be even worse by bloating the stack size of one log to 160 bytes or more. If you box it well you're back to a vec basically

crates/dyn-abi/src/ext/event.rs
DaniPopes: I guess that decoding 2 words into 1 (non indexed address) should fail? But currently doesn't, I'm not really sure how to fix this and it's out of scope

DaniPopes: [`ethabi::decode_whole`](https://docs.rs/ethabi/latest/ethabi/fn.decode_whole.html) says it "Fails, if some data left to decode" so I guess that would be enough to fix this

crates/dyn-abi/src/ext/abi.rs
DaniPopes: `validate` set to `true` is the same as calling the `_whole` version of the function, which I didn't know existed prior to this

prestwich (approved):

PR #275, opened 2023-09-19
chore: do not implement SolType for SolStruct generically

### LINKED ISSUE #36: Move blanket `SolType for SolStruct` impl to `sol!` macro
This code:

https://github.com/ethers-rs/core/blob/23962043923bd859881e1c0021822641763e1564/crates/abi/src/types/struct.rs#L108-L112

DaniPopes:
## Motivation

Closes #36

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


crates/dyn-abi/src/eip712/resolver.rs
```diff
@@ -316,7 +316,8 @@ impl Resolver {
 
     /// Ingest a sol struct typedef.
     pub fn ingest_sol_struct<S: SolStruct>(&mut self) {
-        self.ingest_string(S::eip712_encode_type()).unwrap();
+        self.ingest_string(<S as SolStruct>::eip712_encode_type())
```
DaniPopes, line 319: Both `SolType` and `SolStruct` have a `fn eip712_encode_type`

gakonst (commented):
not following what's made easier here, is it applying less restrictions on the type?

crates/dyn-abi/src/eip712/resolver.rs
prestwich: while we're making a large breaking change to the encode interface, lets also fix this by differentiating the names

DaniPopes:
@gakonst Removing the generic impl makes the compiler give cleaner error messages, fixes some type ambiguity errors I got while trying to add methods to the Encodable trait

crates/dyn-abi/src/eip712/resolver.rs
gakonst: what should the names be?

prestwich: Do we need them on both interfaces, or can we have `eip712_encode_type` on only `SolType` @DaniPopes

DaniPopes: `SolType`'s returns `Some` only for structs, so it should only be on `SolStruct`

DaniPopes: Removed in 820ee0750ada1ed4b35ed6bea7236d71e7aeb653

prestwich (approved):

PR #282, opened 2023-09-19
feat: add support for `rand`

DaniPopes:


crates/primitives/Cargo.toml
```diff
@@ -38,6 +38,9 @@ serde = { workspace = true, optional = true }
 # getrandom
 getrandom = { workspace = true, optional = true }
 
+# rand
+rand = { workspace = true, optional = true, features = ["getrandom"] }
```
gakonst, line 42: do we need to enable some feature for wasm here?

gakonst (approved):
nice

PR #304, opened 2023-09-25
hash_message

supernovahs:
Ref https://github.com/paradigmxyz/reth/pull/4758

### DIFF
--- a/crates/primitives/src/utils.rs
+++ b/crates/primitives/src/utils.rs
@@ -1,4 +1,27 @@
-use crate::bits::FixedBytes;
+use crate::{bits::FixedBytes, B256};
+
+/// Hash a message according to [EIP-191] (version `0x01`).
+///
+/// The final message is a UTF-8 string, encoded as follows:
+/// `"\x19Ethereum Signed Message:\n" + message.length + message`
+///
+/// This message is then hashed using [Keccak-256](keccak256).
+///
+/// [EIP-191]: https://eips.ethereum.org/EIPS/eip-191
+pub fn hash_message<T: AsRef<[u8]>>(message: T) -> B256 {
+    const PREFIX: &str = "\x19Ethereum Signed Message:\n";
+
+    let message = message.as_ref();
+    let len = message.len();
+    let len_string = len.to_string();
+
+    let mut eth_message = Vec::with_capacity(PREFIX.len() + len_string.len() + len);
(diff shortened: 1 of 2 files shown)


crates/primitives/src/utils.rs
```diff
@@ -1,4 +1,27 @@
...
+
+/// Hash a message according to [EIP-191] (version `0x01`).
+///
+/// The final message is a UTF-8 string, encoded as follows:
+/// `"\x19Ethereum Signed Message:\n" + message.length + message`
+///
+/// This message is then hashed using [Keccak-256](keccak256).
+///
+/// [EIP-191]: https://eips.ethereum.org/EIPS/eip-191
+pub fn hash_message<T: AsRef<[u8]>>(message: T) -> B256 {
```
prestwich, line 11: can we prefix this with `eip191_` to match the eip712 methods in sol-types?

crates/primitives/src/utils.rs
```diff
@@ -1,4 +1,27 @@
...
+    const PREFIX: &str = "\x19Ethereum Signed Message:\n";
```
prestwich, line 12: move const out of function to module scope. Rename to `EIP191_PREFIX`

crates/primitives/src/utils.rs
```diff
@@ -1,4 +1,27 @@
...
+    let len = message.len();
+    let len_string = len.to_string();
+
+    let mut eth_message = Vec::with_capacity(PREFIX.len() + len_string.len() + len);
+    eth_message.extend_from_slice(PREFIX.as_bytes());
+    eth_message.extend_from_slice(len_string.as_bytes());
+    eth_message.extend_from_slice(message);
+
+    keccak256(&eth_message)
+}
```
prestwich, line 24: please add the test from ethers-rs

```
// test vector taken from:
    // https://web3js.readthedocs.io/en/v1.2.2/web3-eth-accounts.html#hashmessage
    #[test]
    fn test_hash_message() {
        let hash = hash_message("Hello World");

        assert_eq!(
            hash,
            "a1de988600a42c4b4ab089b619297c17d53cffae5d5120d82d8a92d0bb3b78f2".parse().unwrap()
        );
    }
```

prestwich (changes requested):
good contribution, thanks. a few  changes :)

DaniPopes (approved):

PR #308, opened 2023-09-26
test: move `decode_revert_reason` to alloy and add tests

alessandromazza98:
Related to paradigmxyz/reth#4789

I moved `decode_revert_reason` function here and also add tests:

1. first test is for properly decoding a solidity revert reason
2. second one is properly decoding a random revert reason (should fit for vyper revert reasons)
3. last one to test that too short reason fail since `out.len() < Selector::len_bytes()`

Actually the last test makes me wonder if it is ok to put that `if` sentence in the first line of the function as it could happen that a vyper revert reason is actually shorter than 4 bytes but still valid.

Let me know WDYT.

Also, before I can delete this function from Reth, this PR needs to be merged. Otherwise in Reth there are errors because `decode_revert_reason` does not have any definition.

### DIFF
--- a/crates/sol-types/src/types/error.rs
+++ b/crates/sol-types/src/types/error.rs
@@ -409,8 +409,7 @@ impl PanicKind {
 }
 
 /// Returns the revert reason from the given output data, if it's an abi encoded
-/// String. Returns `None` if the output is not long enough to contain a
-/// function selector or the content is not a valid abi encoded String.
+/// String. Returns `None` if the content is not a valid abi encoded String.
 pub fn decode_revert_reason(out: &[u8]) -> Option<String> {
     // Try to decode as a generic contract error.
     if let Ok(error) = GenericContractError::decode(out, true) {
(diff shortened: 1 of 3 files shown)


crates/sol-types/src/types/error.rs
```diff
@@ -408,6 +408,31 @@ impl PanicKind {
...
+/// Returns the revert reason from the given output data, if it's an abi encoded
+/// String. Returns `None` if the output is not long enough to contain a
+/// function selector or the content is not a valid abi encoded String.
+///
+/// **Note:** it's assumed the `out` buffer starts with the call's signature
+pub fn decode_revert_reason(out: &[u8]) -> Option<String> {
+    // Ensure the output data is long enough to contain a function selector.
+    if out.len() < Selector::len_bytes() {
+        return None
+    }
```
DaniPopes, line 420: I think it's better if we remove this for the reasons you mentioned

DaniPopes (changes requested):

alessandromazza98:
Sorry for lots of commits. I can squash them if it's needed

alessandromazza98:
Sorry for lots of commits. I can squash them if it's needed. Should be ok now.

PR #309, opened 2023-09-27
DynSolEvent & ResolveSolEvent

prestwich:
based on #271 

## Motivation

- Make clear checked/unchecked API for `Log` to allow safe usage as well as instantiating invalid logs
- Add `DynSolEvent`
- Add `ResolveSolEvent` to allow future reuse by non-JSON (e.g. `impl ResolveSolEvent for syn_solidity::ItemEvent` would be cool)

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


Evalir (commented):
ACK—quite like the checked/unchecked API myself.

prestwich:
should the unchecked API be `unsafe`?  e.g. `unsafe fn new_unchecked` ?

DaniPopes:
My stance on that is if an invalid value causes either UB, or panics everywhere to the point it's unusable. If it only causes a panic in small sections or not at all, then it should not be marked as unsafe. I don't think this is the case here

prestwich:
cool. we'll leave them as safe then

crates/dyn-abi/src/event.rs
```diff
@@ -0,0 +1,204 @@
...
+        let log = Log::new_unchecked(
+            vec![
+                t0,
+                b256!("0000000000000000000000000000000000000000000000000000000000012321"),
+            ],
+            bytes!(
+                "
+			0000000000000000000000000000000000000000000000000000000000012345
+			0000000000000000000000000000000000000000000000000000000000054321
+			"
```
gakonst, line 184: nit de indent

crates/dyn-abi/src/event.rs
```diff
@@ -0,0 +1,204 @@
...
+            body: DynSolType::Tuple(vec![DynSolType::Tuple(vec![
+                DynSolType::Address,
+                DynSolType::Address,
+            ])]),
+        };
+
+        let decoded = event.decode_log(&log, true).unwrap();
+        assert_eq!(
+            decoded.indexed,
+            vec![DynSolValue::Address(address!(
```
gakonst, line 199: maybe eventually we add `sol_address!` as extra sugar, but might be overkill

crates/dyn-abi/src/event.rs
```diff
@@ -0,0 +1,204 @@
...
+            match topics.size_hint() {
+                (n, Some(m)) if n == m && n != num_topics => {
+                    return Err(Error::TopicLengthMismatch {
+                        expected: num_topics,
+                        actual: n,
+                    })
+                }
+                _ => {}
+            }
+        }
```
gakonst, line 69: 👍

gakonst (approved):

crates/dyn-abi/src/ext/event.rs
```diff
@@ -1,15 +1,50 @@
...
+            return Err(Error::TopicLengthMismatch {
+                expected: 4,
+                actual: num_topics,
+            })
+        }
+
+        Ok(DynSolEvent::new_unchecked(
+            topic_0,
+            indexed,
+            DynSolType::Tuple(body),
```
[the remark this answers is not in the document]
prestwich: well, it's really just a sequence of types. but Tuple is the way to represent that. could be some cleanup here tho

crates/dyn-abi/src/event.rs
prestwich: this is @DaniPopes code I just rearranged it

PR #310, opened 2023-09-27
feat: add human-readable ABI parser

alessandromazza98:
Closes #291 

This is a draft just to let you know how I'm trying to solve it @DaniPopes.

Right now I implemented a `parse` method only for `Error` but the code is easily re-usable.
Also now all my code is inside `item.rs` but I can move most of the real parser logic inside `alloy-sol-type-parser` as you suggest in the issue description (maybe in a new file inside of that?)

I also add some tests so that you can see how it works. Also these tests are in the `item.rs` file but I can move them somewhere else.

PS: of course the CI gives error because I used `println!` in the tests just to have a quick view of how the parsing work. Later I can delete them or change them with the use of `assert`

PR #329, opened 2023-10-03
feat: add parsing support for JSON items

### LINKED ISSUE #291: [Feature] Implement human-readable ABI parser
### Component

dyn-abi, json-abi, Other (please provide more details)

- add `parse` methods to `alloy_json_abi::{Error, Event, Function}` that parses `$name($($inputs),*)($($outputs),*)` into these structs (outputs is just for `Function`)
- implement parser in `alloy-sol-type-parser`

### Additional context

_No response_

DaniPopes:
Add `FromStr` and `parse` associated functions to `json-abi` items and params.

Supersedes #310
Closes #291

### DIFF
--- a/crates/json-abi/src/abi.rs
+++ b/crates/json-abi/src/abi.rs
@@ -111,7 +111,7 @@ impl JsonAbi {
         serde_json::from_str(json)
     }
 
-    /// Loads contract from a JSON [Reader](std::io::Reader).
+    /// Loads contract from a JSON [Reader](std::io::Read).
     ///
     /// This is a convenience wrapper around [`serde_json::from_str`].
     #[cfg(all(feature = "std", feature = "serde_json"))]
--- a/crates/json-abi/src/abi.rs
+++ b/crates/json-abi/src/abi.rs
@@ -1,10 +1,7 @@
 use crate::{AbiItem, Constructor, Error, Event, Fallback, Function, Receive};
-use alloc::{
-    collections::{btree_map, btree_map::Values},
-    string::String,
-    vec::Vec,
-};
+use alloc::{collections::btree_map, string::String, vec::Vec};
 use alloy_primitives::Bytes;
+use alloy_sol_type_parser::{Error as ParserError, Result as ParserResult};
 use btree_map::BTreeMap;
 use core::{fmt, iter, iter::Flatten};
 use serde::{
@@ -12,6 +9,28 @@ use serde::{
     ser::SerializeSeq,
     Deserialize, Deserializer, Serialize,
 };
+
+macro_rules! set_if_none {
+    ($opt:expr, $val:expr) => { set_if_none!(stringify!($opt) => $opt, $val) };
+    (@serde $opt:expr, $val:expr) => { set_if_none!(serde::de::Error::duplicate_field(stringify!($opt)) => $opt, $val) };
+    ($name:expr => $opt:expr, $val:expr) => {{
+        if $opt.is_some() {
+            return Err($name)
+        }
+        $opt = Some($val);
+    }};
+}
+
+macro_rules! entry_and_push {
+    ($map:expr, $v:expr) => {
+        $map.entry($v.name.clone())
+            .or_default()
+            .push($v.into_owned())
+    };
+}
+
+type FlattenValues<'a, V> = Flatten<btree_map::Values<'a, String, Vec<V>>>;
+type FlattenIntoValues<V> = Flatten<btree_map::IntoValues<String, Vec<V>>>;
 
 /// The JSON contract ABI, as specified in the [Solidity ABI spec][ref].
 ///
@@ -32,6 +51,16 @@ pub struct JsonAbi {
     pub errors: BTreeMap<String, Vec<Error>>,
 }
 
+impl<'a> FromIterator<AbiItem<'a>> for JsonAbi {
+    fn from_iter<T: IntoIterator<Item = AbiItem<'a>>>(iter: T) -> Self {
+        let mut abi = Self::new();
+        for item in iter {
+            let _ = abi.insert_item(item);
+        }
+        abi
+    }
+}
+
 impl JsonAbi {
     /// Creates an empty ABI object.
     #[inline]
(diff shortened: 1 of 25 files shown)


crates/json-abi/src/item.rs
```diff
@@ -356,7 +358,95 @@ impl AbiItem<'_> {
...
+    #[inline]
+    pub fn signature(&self) -> String {
+        signature("constructor", &self.inputs, None)
+    }
+
+    /// Computes this constructor's selector: `keccak256(self.signature())[..4]`
+    #[inline]
+    pub fn selector(&self) -> Selector {
+        selector(&self.signature())
+    }
```
DaniPopes, line 415: Idk how useful this is on `Constructor`s

crates/sol-type-parser/src/parameter.rs
```diff
@@ -0,0 +1,202 @@
...
+        trace(
+            "ParameterSpecifier",
+            spanned(|input: &mut &'a str| {
+                let ty = TypeSpecifier::parser(input)?;
+                let mut indexed = false;
+                let mut name = opt_ws_ident(input)?;
+                // TODO: ?
+                // if let Some("storage" | "memory" | "calldata") = name {
+                //     name = opt_ws_ident(input)?;
+                // }
```
DaniPopes, line 44: Should we add this?

gakonst: I wonder if e.g. you were building Stylus maybe you'd need this to differentiate between when to copy calldata vs not, but I think we can leave as a todo? 

alternatively it might be a simple matter of having an enum called `pointer: Storage | Memory | Calldata` as part of the struct like you did with the indexed events?

crates/sol-type-parser/src/lib.rs
```diff
@@ -83,3 +93,55 @@ pub(crate) fn str_parser<'a, E: ParserError<&'a str> + AddContext<&'a str, StrCo
...
+    )
+}
+
+pub(crate) fn opt_ws_ident<'a>(input: &mut &'a str) -> PResult<Option<&'a str>> {
+    preceded(space0, opt(parse_identifier)).parse_next(input)
+}
+
+// Not public API.
+#[doc(hidden)]
+pub fn __internal_parse_signature<'a, const OUT: bool, F: Fn(ParameterSpecifier<'a>) -> T, T>(
```
gakonst, line 121: parsing code is always so cursed, think we could use docs in these internally for contributors, but no action required now

crates/sol-type-parser/src/error.rs
```diff
@@ -57,12 +57,12 @@ impl Error {
...
     #[cold]
     fn new_(s: &str, e: &dyn fmt::Display) -> Self {
-        Self(Repr(format!("{s}{e}")))
+        Self(Repr(format!("{s}{e}").into_boxed_str()))
     }
 }
 
 #[derive(Clone, PartialEq, Eq)]
-struct Repr(String);
+struct Repr(Box<str>);
```
gakonst, line 65: why do we need this in practice?

crates/json-abi/src/item.rs
```diff
@@ -373,7 +463,63 @@ impl Error {
...
+    ///
+    /// ```
+    /// # use alloy_json_abi::{Function, Param, StateMutability};
+    /// assert_eq!(
+    ///     Function::parse("toString(uint number)(string s)"),
+    ///     Ok(Function {
+    ///         name: "toString".to_string(),
+    ///         inputs: vec![Param::parse("uint number").unwrap()],
+    ///         outputs: vec![Param::parse("string s").unwrap()],
+    ///         state_mutability: StateMutability::Payable,
```
gakonst, line 509: Should we default to payable or non payable? I think non-payable?

crates/json-abi/src/item.rs
```diff
@@ -373,7 +463,63 @@ impl Error {
...
+    ///         outputs: vec![],
+    ///         state_mutability: StateMutability::Payable,
+    ///     }),
+    /// )
+    /// ```
+    ///
+    /// [Function]s also support parsing output tuples:
+    ///
+    /// ```
+    /// # use alloy_json_abi::{Function, Param, StateMutability};
```
gakonst, line 502: This example says that the output is a tuple but the output is a string

crates/json-abi/src/item.rs
```diff
@@ -373,7 +463,63 @@ impl Error {
...
+    /// assert_eq!(
+    ///     Function::parse("toString(uint number)(string s)"),
```
gakonst, line 504: Do the parsers still work if the string starts with the word `function` for `Function::parse` and `event` for `Event::parse`?

crates/json-abi/src/item.rs
```diff
@@ -356,7 +358,95 @@ impl AbiItem<'_> {
...
+    type Err = ParserError;
+
+    #[inline]
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::parse(s)
+    }
+}
+
+impl Constructor {
+    /// Parses a Solidity event signature string:
```
gakonst, line 371:
```suggestion
    /// Parses a Solidity constructor string:
```

Constructor does not take anonymous

crates/json-abi/src/item.rs
gakonst: am indifferent, cc @prestwich

gakonst (approved):
supportive

crates/sol-type-parser/src/error.rs
DaniPopes: Not needed really, just want to abstract it away from the main struct. If you mean `Box<str>`, it's one word smaller than `String` (24 -> 16)

crates/json-abi/src/item.rs
DaniPopes: Payable is the EVM "default", the rest are higher level abstractions

crates/json-abi/src/item.rs
prestwich: not useful at all. constructors are never called as functions

crates/json-abi/src/item.rs
DaniPopes: Right, I meant it as an additional parameter list for the returns, after the arguments one

crates/json-abi/src/item.rs
DaniPopes: No, that should be handled prior. I'll make that clearer, and add a function to `AbiItem` as a way to handle it

crates/json-abi/src/item.rs
prestwich: non-payable is default

DaniPopes: In Solidity yes, you have to opt out

prestwich (approved):
approving for when you feel it's ready @DaniPopes

crates/json-abi/src/item.rs
DaniPopes: I guess that makes more sense because we're parsing Solidity strings after all

crates/json-abi/src/item.rs
DaniPopes: Done in c947ff7e7a361f32d396d89706a8e4afa8c46d84

crates/json-abi/src/item.rs
DaniPopes: Updated to say output parameters

DaniPopes:
Small thing about this is that all parse functions use the same parser, so they will accept and ignore stuff like "indexed" and state mutability instead of returning an error, which I don't think is that bad

crates/sol-type-parser/src/parameter.rs
DaniPopes: Added in 9be37b13050d1a3d246c873bb3f7a87354dd1f5e

crates/json-abi/src/abi.rs
```diff
@@ -32,22 +51,69 @@ pub struct JsonAbi {
...
+    ///     "constructor(string symbol, string name)",
+    ///     "function transferFrom(address from, address to, uint value)",
+    ///     "function balanceOf(address owner)(uint balance)",
+    ///     "event Transfer(address indexed from, address indexed to, address value)",
+    ///     "error InsufficientBalance(account owner, uint balance)",
+    ///     "function addPerson(tuple(string, uint16) person)",
+    ///     "function addPeople(tuple(string, uint16)[] person)",
+    ///     "function getPerson(uint id)(tuple(string, uint16))",
+    ///     "event PersonAdded(uint indexed id, tuple(string, uint16) person)",
+    /// ])?;
```
gakonst, line 91: I believe we also allowed doing in-line definition of a struct `struct Person { string name; uint16 age }` and then your `addPerson` function would be `addPerson(Person person)` instead of the tuple above. Does this also work if the word `tuple` is missing?

crates/json-abi/src/item.rs
```diff
@@ -207,7 +193,58 @@ impl<'a, 'r> From<&'r AbiItem<'a>> for &'r private::AbiItem<'a> {
...
+        <&private::AbiItem<'_>>::from(self).serialize(serializer)
+    }
+}
+
+impl<'de: 'a, 'a> Deserialize<'de> for AbiItem<'a> {
+    #[inline]
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        private::AbiItem::deserialize(deserializer).map(Into::into)
+    }
+}
```
gakonst, line 208: I forget, why do we have the `private` module?

gakonst (approved):

crates/json-abi/src/item.rs
DaniPopes: Comment at the top of the file

crates/json-abi/src/abi.rs
DaniPopes: 1. that's out of scope of this, it parses items one at a time
2. tuple is an optional keyword

crates/json-abi/src/item.rs
DaniPopes: Well, dug into this a bit more and we can get rid of it. #330

PR #330, opened 2023-10-04
refactor(json-abi): remove duplicate item serde hack

DaniPopes:
Based on #329

Simplifies serde for items and `AbiItem` by using `tag = "type"` on the item structs directly.
Perf is slightly (2-5%) better for ser, while unchanged for de


gakonst (approved):
very nice

PR #333, opened 2023-10-05
feat(sol-types): introduce `SolValue`, make `Encodable` an impl detail

DaniPopes:
- Renamed existing `Encodable` to `SolTypeValue` and make it clear that it and `SolType` should not be implemented manually
- Add a new `SolValue` trait with the useful methods from `SolType`, but instead of `rust : &Self::RustType` they take receiver self
- `SolValue` is implemented for all ABI Rust types and just delegates everything to `SolTypeEncodable` when `Self: SolTypeEncodable`
- Moved implementations of previous `&Self::RustType` methods to `SolTypeEncodable` so they can be used in `Encodable`


crates/sol-types/src/types/encodable.rs
```diff
@@ -0,0 +1,405 @@
...
+    {
+        Self::SolType::abi_encode_params(self)
+    }
+
+    /// ABI-decode this type from the given data.
+    ///
+    /// See [`SolType::abi_decode`] for more information.
+    fn abi_decode(data: &[u8], validate: bool) -> Result<<Self::SolType as SolType>::RustType> {
+        Self::SolType::abi_decode(data, validate)
+    }
```
DaniPopes, line 125: I don't think there's a way to say `where Self = Self::SolType::RustType` so for example `str::abi_decode` will return `String` here, see tests below

crates/sol-types/src/types/encodable.rs
```diff
@@ -0,0 +1,405 @@
...
+    // Make sure these are in scope
+    #[allow(unused_imports)]
+    use crate::{SolType as _, SolTypeEncodable as _};
+
+    #[test]
+    fn inference() {
+        false.sol_type_name();
+        // false.abi_encoded_size();
+        // false.eip712_data_word();
+        // false.abi_encode_packed_to(&mut vec![]);
```
DaniPopes, line 244: These commented out methods are both in `SolTypeEncodable` and `Encodable`. Do we care enough to rename one of them?

crates/sol-types/src/types/encodable.rs
prestwich: what about using from? 

`where Self: From<Self::SolType::RustType>`

crates/sol-types/src/types/encodable.rs
prestwich: given that users are discouraged from importing `SolTypeEncodable` , i would skip the rename

prestwich: i guess this might make some difficulty for like, rust-analyzer auto-hints for imports? or for compiler messages?

prestwich: so i have revised my opinion in the last 2 minutes let's rename the internal ones, and keep the clear names for the user-facing one

DaniPopes:
Also is there a better name than `Encodable`? This trait is a helper for everything `SolType` does, not just encoding

crates/sol-types/src/types/event/mod.rs
```diff
@@ -13,7 +13,7 @@ pub use topic_list::TopicList;
 
 /// Solidity event.
 ///
-/// ### Implementer's Guide
+/// # Implementer's Guide
 ///
 /// We do not recommend implementing this trait directly. Instead, we recommend
```
prestwich: wording should match updated wording from other modules?

crates/sol-types/src/eip712.rs
```diff
@@ -157,13 +154,11 @@ impl Eip712Domain {
...
 
         out.reserve(self.abi_encoded_size());
-        let name = self.name.as_ref().map(cow_keccak256);
-        encode_opt::<sol_data::FixedBytes<32>, _>(name.as_ref(), out);
-        let version = self.version.as_ref().map(cow_keccak256);
-        encode_opt::<sol_data::FixedBytes<32>, _>(version.as_ref(), out);
-        encode_opt::<sol_data::Uint<256>, _>(self.chain_id.as_ref(), out);
-        encode_opt::<sol_data::Address, _>(self.verifying_contract.as_ref(), out);
-        encode_opt::<sol_data::FixedBytes<32>, _>(self.salt.as_ref(), out);
+        encode_opt!(self.name.as_ref().map(cow_keccak256));
```
prestwich, line 157: so clean 👀

prestwich (approved):
🤩

PR #342, opened 2023-10-08
fix: json-abi not using anonymous when converting to interface

Will-Smith11:
Hey so was seeing what would happen if I ran all of the abis ever verified on etherscan through alloy's `sol` macro and this was one of the errors I ran into.
This error occurs due to the fact that when we go to convert parsed json-abi into a interface for syn-solidity to then parse as a normalization step, the anonymous keyword on events was not getting included into the interface and was causing an error.
Heres an example:
https://etherscan.io/address/0x29c1aaa2e5bd43692f516d9a0c3d2a0f37b59053#code
image

### DIFF
--- a/crates/json-abi/src/to_sol.rs
+++ b/crates/json-abi/src/to_sol.rs
@@ -327,7 +327,7 @@ impl<IN: ToSol> ToSol for AbiFunction<'_, IN> {
         }
 
         if self.anonymous {
-            out.push_str("anonymous");
+            out.push_str(" anonymous");
         }
 
         out.push(';');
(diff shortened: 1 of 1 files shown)


crates/json-abi/src/to_sol.rs
```diff
@@ -320,6 +326,10 @@ impl<IN: ToSol> ToSol for AbiFunction<'_, IN> {
             out.push(')');
         }
 
+        if self.anonymous {
+            out.push_str("anonymous");
```
DaniPopes, line 330:
```suggestion
            out.push_str(" anonymous");
```

DaniPopes (approved):
Nice catch, thanks!

PR #350, opened 2023-10-08
feat: support legacy json abis

Will-Smith11:
currently there is no support for legacy abis. I quickly added it, ignore the extra commits, from my fork where i also quickly patched rust keyword impl by using r#


crates/json-abi/src/item.rs
```diff
@@ -9,19 +9,17 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 macro_rules! abi_items {
     ($(
         $(#[$attr:meta])*
-        $vis:vis struct $name:ident : $name_lower:literal {$(
+        $vis:vis struct $name:ident : $($name_lower:tt),+ {$(
             $(#[$fattr:meta])*
             $fvis:vis $field:ident : $type:ty,
         )*}
     )*) => {
+        trait FlattenStateMutability {
```
DaniPopes, line 17: I think we can simplify this and flatten at serde-time in `StateMutability`. There's also `constant: bool` alongside `payable: bool`

Will-Smith11: does `constant: bool` effect the generation of the bindings at all? I totally agree with flattening it but do you think we can just ignore the `constant` field?

DaniPopes: I assume it's converted like this
```rust
match (constant, payable) {
    (false, false) => StateMutability::NonPayable,
    (false, true) => StateMutability::Payable,
    (true, false) => StateMutability::View,
    (true, true) => StateMutability::Payable, // I don't think this can happen
}
```

prestwich: `(true, true)` should be an error, no? as constant is a contract not to read or write state, both being set would be an invalid ABI?

Will-Smith11:
hey sorry was super busy with work. made the changes yall wanted

gakonst:
@DaniPopes @prestwich what do we want to do with this PR? totally defer to you guys

DaniPopes:
I don't think we want this

PR #375, opened 2023-10-18
fix: fix some typos

shuoer86:
fix typos in the file 'crates/syn-solidity/tests/contracts/Solady.sol' forked from https://github.com/vectorized/solady which is already fixed.


prestwich:
Given that this contract is vendored from a 3rd party, we would like to keep it as close to the original state as possible, that includes preserving typos

shuoer86:
That's all right.

PR #380, opened 2023-10-20
feat(dyn-abi): `DynSolType::coerce_str`

DaniPopes:
## Motivation

Port of [`ethabi::Tokenizer`](https://docs.rs/ethabi/latest/ethabi/token/trait.Tokenizer.html), with extra features

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/dyn-abi/src/coerce.rs
+++ b/crates/dyn-abi/src/coerce.rs
@@ -1,12 +1,14 @@
 use crate::{ty::as_tuple, DynSolType, DynSolValue, Result};
-use alloy_primitives::{Address, FixedBytes, Function, Sign, B256, I256, U256};
+use alloy_primitives::{Address, FixedBytes, Function, Sign, I256, U256};
 use alloy_sol_type_parser::utils::{array_parser, char_parser, spanned};
+use alloy_sol_types::Word;
 use core::fmt;
 use hex::FromHexError;
 use winnow::{
-    ascii::{alpha0, alpha1, hex_digit1, space0},
-    combinator::{delimited, dispatch, fail, opt, preceded, success},
-    error::{ContextError, ErrMode, ErrorKind, FromExternalError},
+    ascii::{alpha0, alpha1, digit1, hex_digit1, space0},
+    combinator::{cut_err, dispatch, fail, opt, preceded, success},
+    error::{ContextError, ErrMode, ErrorKind, FromExternalError, ParserError, StrContext},
+    token::take_while,
     trace::trace,
     PResult, Parser,
 };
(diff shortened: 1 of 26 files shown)


crates/dyn-abi/src/coerce.rs
```diff
@@ -0,0 +1,861 @@
...
+    let Some(delim) = input.chars().next() else {
+        return Ok("")
+    };
+    let has_delim = matches!(delim, '"' | '\'');
+    if has_delim {
+        *input = &input[1..];
+    }
+
+    // TODO: escapes?
+    let min = if has_delim { 0 } else { 1 };
```
mattsse, line 328: hmm, good question, maybe this could happen with same escaped json inputs, but I think we can ignore that

mattsse (approved):
amazing work!

it's been a while since I read some nom code but I still love how readable this is.

test coverage looks decent.

this will simplify so much

crates/dyn-abi/src/coerce.rs
```diff
@@ -0,0 +1,861 @@
...
+fn uint_units(input: &mut &str) -> PResult<usize> {
+    alpha0(input).and_then(|s| {
+        Ok(match s {
+            "ether" => 18,
+            "gwei" | "nano" | "nanoether" => 9,
+            "" | "wei" => 0,
+            _ => return Err(ErrMode::from_error_kind(input, ErrorKind::Fail)),
+        })
+    })
+}
```
Evalir, line 263: This is super handy, lets us replace a few utilities we have in foundry

wdyt about adding more units?

crates/dyn-abi/src/coerce.rs
```diff
@@ -0,0 +1,861 @@
...
+
+        if cfg!(feature = "std") {
+            assert_eq!(
+                DynSolType::Uint(256).coerce_str("0.01 ether").unwrap(),
+                DynSolValue::Uint(U256::from_str("10000000000000000").unwrap(), 256)
+            );
+
+            assert_eq!(
+                DynSolType::Uint(256)
+                    .coerce_str("0.000000000000000001ether")
```
Evalir, line 606: nice

Evalir (approved):
Super readable! This is gonna be super handy—love it

crates/dyn-abi/src/coerce.rs
DaniPopes: Do you have anything in mind? Solidity only supports `wei`, `gwei`, and `ether`

prestwich (commented):
do we have documentation of all the formats supported?

DaniPopes:
Added in 0e974a13502f264c4ded94d36395123b3fe628e0 @prestwich

prestwich:
perfection

PR #409, opened 2023-11-09
Expand coerce string logic

klkvr:
## Motivation

Right now string parsing by default (if string is not surrounded by parentheses) can't parse strings which contain one of `,`, `]` or `)`. This happens because the same parser is used when parsing plain strings, arrays of strings and tuples containing strings.

For example, following coercings will fail:
```
a, b

a (b)

a [b]
```

Such behaviour results in [this foundry issue](https://github.com/foundry-rs/foundry/issues/6265)

## Solution

Create and use separate parsers for string parsing in scope of tuples and arrays. For tuples parser stops at `,` and `)`. For arrays at `,` and `]`

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/dyn-abi/src/coerce.rs
+++ b/crates/dyn-abi/src/coerce.rs
@@ -103,6 +103,24 @@ impl DynSolType {
             as_tuple!(Self tys) => tuple(tys).parse_next(input).map(DynSolValue::Tuple),
         })
     }
+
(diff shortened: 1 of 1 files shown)


DaniPopes:
Hi, sorry I was already working on a fix before you posted this. I'll add you as an author to #410.

Issue #402, opened 2023-11-04
[Feature] Derive Hash for `DynSolType` 


brockelmore:
### Component

dyn-abi

### Describe the feature you would like

Feels like there isn't a reason for Hash to not be derived on `DynSolType`, so we should derive it (and I need Hash on DynSolType). Likely there are other types that similarly miss the hash derivation. 

My defaults are `Clone, Debug, PartialEq, Eq, PartialOrd, Ord, and Hash` for all types if possible (`PartialOrd` and `Ord` only if they make sense) + `Copy` for types that are copy-able. This isn't the issue for double checking the derive on all the types but something to keep in mind.

### Additional context

Relevant Source: https://github.com/alloy-rs/core/blob/e76158bb83644e042c4e55c4a382aeb389ae748b/crates/dyn-abi/src/ty.rs#L83

PR #411, opened 2023-11-14
feat: Added Hash to DynSolType and StructProp

NianSol:
I saw the open issue #402 and decided to implement it

## Solution

Simply added hash to the structs and tests to make sure nothing broke

## PR Checklist

- [] Added Tests - No, it's a default crate feature
- [?] Added Documentation - Not sure if there was a need
- [] Breaking changes - No

### DIFF
--- a/crates/dyn-abi/src/ty.rs
+++ b/crates/dyn-abi/src/ty.rs
@@ -18,7 +18,7 @@ macro_rules! as_tuple {
 }
 pub(crate) use as_tuple;
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 struct StructProp {
     name: String,
     ty: DynSolType,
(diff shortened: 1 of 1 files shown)


crates/dyn-abi/src/ty.rs
```diff
@@ -937,4 +937,41 @@ mod tests {
...
     }
+    #[test]
+    fn test_hash_implementation() {
+        // Create instances for DynSolType
+        let int_42 = DynSolType::Int(42);
+        let int_43 = DynSolType::Int(43);
+        let bool_type = DynSolType::Bool;
+
+        // Create instances for StructProp
+        let prop_a = StructProp { name: "prop".to_string(), ty: int_42.clone() };
```
DaniPopes, line 948: you'll need to import alloc::string::ToString

DaniPopes (changes requested):

crates/dyn-abi/src/ty.rs
NianSol: ok, i'll do it later today and send another pr

NianSol:
Closed while I search for an alternative to STD default hasher for the tests

NianSol:
Closed while I search for an alternative to STD default hasher for the tests

DaniPopes (approved):
Thanks!

PR #413, opened 2023-11-15
fix: wrong as_u8 generated for enum

### LINKED ISSUE #412: [Bug] as_u8 enum generated from sol! is incorrect
### Component

sol-types, sol! macro

0.4.2

### Operating System

macOS (Apple Silicon)

### Describe the bug

I am using sol! macro to generate enum to use it for eip712 signing but when see the hash struct it is wrong. After investigate, I found that the as_u8 is wrong as i tested on rust playground 
```
#[derive(Clone, Copy)]
#[repr(u8)]
enum Color {
    Red = 0,
    Green = 1,
    Blue = 2,
}

impl Color {
    fn as_u8(&self) -> &u8 {
        unsafe { &*::core::ptr::addr_of!(self).cast::<u8>() }
    }
}

fn main() {
    let color = Color::Red;
    let u8_value = color.as_u8();
    println!("The u8 value of the color is: {}", u8_value);
}
```

For this code it will print something like "The u8 value of the color is: 159"

huyhduong1401:
## Motivation
I found a bug which I open an issue in https://github.com/alloy-rs/core/issues/412

Closes #412

## Solution
Directly casting pointer type

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro/src/expand/enum.rs
+++ b/crates/sol-macro/src/expand/enum.rs
@@ -193,7 +193,7 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, enumm: &ItemEnum) -> Result<TokenStream> 
                 #[allow(unsafe_code, clippy::inline_always)]
                 #[inline(always)]
                 fn as_u8(&self) -> &u8 {
-                    unsafe { &*::core::ptr::addr_of!(self).cast::<u8>() }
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
Thanks!

PR #417, opened 2023-11-18
chore: rename `TokenType` GAT and trait to `Token`

DaniPopes:
Last few remaining breaking changes before 0.5.


prestwich:
this is a good rename

PR #419, opened 2023-11-18
feat: enable ruint ssz when primitives ssz

prestwich:
Enables ruint ssz feature when alloy-primitives SSZ feature is enabled

Blocked by release of ruint@1.11.1 https://github.com/recmo/uint/pull/345

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -71,7 +71,7 @@ getrandom = ["dep:getrandom"]
 rand = ["dep:rand", "getrandom", "ruint/rand"]
 rlp = ["dep:alloy-rlp", "ruint/alloy-rlp"]
 serde = ["dep:serde", "bytes/serde", "hex/serde", "ruint/serde"]
-ssz = ["dep:ethereum_ssz", "std"]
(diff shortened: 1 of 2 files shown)


crates/primitives/Cargo.toml
```diff
@@ -71,7 +71,7 @@ getrandom = ["dep:getrandom"]
 rand = ["dep:rand", "getrandom", "ruint/rand"]
 rlp = ["dep:alloy-rlp", "ruint/alloy-rlp"]
 serde = ["dep:serde", "bytes/serde", "hex/serde", "ruint/serde"]
-ssz = ["dep:ethereum_ssz", "std"]
+ssz = ["dep:ethereum_ssz", "std", "ruint/ssz"]
```
DaniPopes, line 74: You should also bump ruint version in the root manifest

DaniPopes (changes requested):

crates/primitives/Cargo.toml
prestwich: 🥇

DaniPopes (approved):

Issue #389, opened 2023-10-27
[Feature] Full function signature parsing


Evalir:
It would be nice for `Function` to be able to parse a full function signature e.g `function foo(uint256 a, uint256 b) external returns (uint256)`—this functionality is required in foundry and human-readable ABI strings fall short for view modifiers.

gakonst:
Can this be closed?

DaniPopes:
This is only missing parsing modifiers like `external`

PR #420, opened 2023-11-18
feat(json-abi): permit keyword prefixes in HR parser

DaniPopes:
## Motivation

Ref https://github.com/alloy-rs/core/issues/389

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-abi/src/utils.rs
+++ b/crates/json-abi/src/utils.rs
@@ -82,7 +82,7 @@ pub(crate) fn parse_maybe_prefixed<F: FnOnce(&str) -> R, R>(
     parser: F,
 ) -> R {
     if let Some(stripped) = s.strip_prefix(prefix) {
-        if stripped.starts_with(|c: char| c.is_ascii_whitespace()) {
+        if stripped.starts_with(char::is_whitespace) {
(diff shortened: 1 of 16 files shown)


crates/json-abi/src/utils.rs
```diff
@@ -72,16 +72,31 @@ pub(crate) fn selector(preimage: &str) -> Selector {
...
+/// Strips `prefix` from `s` before parsing with `parser`. `prefix` must be followed by whitespace.
+pub(crate) fn parse_maybe_prefixed<F: FnOnce(&str) -> R, R>(
+    mut s: &str,
+    prefix: &str,
+    parser: F,
+) -> R {
+    if let Some(stripped) = s.strip_prefix(prefix) {
+        if stripped.starts_with(|c: char| c.is_ascii_whitespace()) {
+            s = stripped.trim_start();
+        }
```
mattsse, line 87: if we need to explicitly handle ascii whitespace,
then this should be `trim_start_matches(is_ascii_whitespace)` I think, or just trim_start directly?

mattsse (approved):
lgtm

PR #424, opened 2023-11-20
feat(primitives): left and right padding conversions

prestwich:
## Motivation

Current `try_from` and `from_slice` requires exact length. These convenience functions provide variants that support shorter slices by left- or right-padding to the correct length

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/fixed.rs
+++ b/crates/primitives/src/bits/fixed.rs
@@ -471,12 +471,12 @@ impl<const N: usize> FixedBytes<N> {
     ///
     /// # Panics
     ///
-    /// If `src.len() > N`
+    /// Panics if `src.len() > N`.
     #[track_caller]
     #[inline]
     pub fn left_padding_from(value: &[u8]) -> Self {
         let len = value.len();
-        assert!(len <= N, "slice is too large. Expected <={} bytes, got {}", N, len);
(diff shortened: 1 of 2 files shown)


crates/primitives/src/bits/fixed.rs
```diff
@@ -462,6 +462,44 @@ impl<const N: usize> FixedBytes<N> {
...
+    /// with zeroes if necessary.
+    ///
+    /// # Note
+    ///
+    /// The given bytes are interpreted in big endian order.
+    ///
+    /// # Panics
+    ///
+    /// If `src.len() > N`
+    pub fn right_padding_from(value: &[u8]) -> Self {
```
DaniPopes, line 495: track caller inline

crates/primitives/src/bits/fixed.rs
```diff
@@ -462,6 +462,44 @@ impl<const N: usize> FixedBytes<N> {
...
+        let len = value.len();
+        assert!(len <= N, "slice is too large. Expected <={} bytes, got {}", N, len);
+        let mut bytes = Self::ZERO;
+        bytes[..len].copy_from_slice(value);
+        bytes
+    }
+
```
DaniPopes, line 502: needs same methods on fixed bytes macro

DaniPopes (changes requested):

PR #425, opened 2023-11-21
fix(sol-macro): correctly print Custom types in parameters

DaniPopes:
## Motivation

Follow-up fix to #422

Previously custom type in parameters would get formatted as "tuple" even when not tuples, like enums.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm

PR #432, opened 2023-11-23
feat(primitives): improve utils

DaniPopes:
## Motivation

Improve existing and add new utilities. Mostly the `units` utilities from ethers-core.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/utils/mod.rs
+++ b/crates/primitives/src/utils/mod.rs
@@ -1,7 +1,8 @@
 use crate::{bits::FixedBytes, B256};
-use alloc::{string::ToString, vec::Vec};
+use alloc::vec::Vec;
+
 /// The prefix used for hashing messages according to EIP-191.
-const EIP191_PREFIX: &str = "\x19Ethereum Signed Message:\n";
+pub const EIP191_PREFIX: &str = "\x19Ethereum Signed Message:\n";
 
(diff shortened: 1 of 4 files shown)


Evalir (approved):

crates/primitives/src/utils/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// Simple interface to the [`Keccak-256`] hash function.
+///
+/// [`Keccak-256`]: https://en.wikipedia.org/wiki/SHA-3
+pub fn keccak256<T: AsRef<[u8]>>(bytes: T) -> B256 {
+    fn keccak256(bytes: &[u8]) -> B256 {
+        let mut output = MaybeUninit::<B256>::uninit();
+
+        cfg_if::cfg_if! {
+            if #[cfg(all(feature = "native-keccak", not(feature = "tiny-keccak")))] {
```
mattsse, line 65: do we want to throw a compile error if both features are used?

mattsse (approved):
lgtm

crates/primitives/src/utils/mod.rs
DaniPopes: No, we already went through this when this was added

PR #437, opened 2023-11-28
fix public leak

kristian1108:
## Motivation

Anvil is currently not building on master. I'm not entirely sure how this crate is building, but would love to learn if someone has an explanation.

## Solution

There's a simple leaked private type error. This fixes it by making the relevant struct public.

## PR Checklist

- [ n/a] Added Tests
- [ n/a] Added Documentation
- [n/a] Breaking changes


DaniPopes:
I think it's still worthwhile to include -- see discussions in <https://t.me/foundry_support/47179>

PR #446, opened 2023-12-05
feat(primitives): update Bytes formatting, add UpperHex

DaniPopes:
## Motivation

Don't print "Bytes()" in Debug

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bytes/mod.rs
+++ b/crates/primitives/src/bytes/mod.rs
@@ -21,21 +21,25 @@ pub struct Bytes(pub bytes::Bytes);
 
 impl fmt::Debug for Bytes {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("Bytes(")?;
-        f.write_str(&self.hex_encode())?;
-        f.write_str(")")
+        fmt::LowerHex::fmt(self, f)
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/primitives/src/bytes/mod.rs
```diff
@@ -21,21 +21,25 @@ pub struct Bytes(pub bytes::Bytes);
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&self.hex_encode())
+        fmt::LowerHex::fmt(self, f)
     }
 }
 
 impl fmt::LowerHex for Bytes {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&self.hex_encode())
+        f.pad(&hex::encode_prefixed(self.as_ref()))
```
prestwich, line 36: we should match std behavior here by adding `0x` only  when `fmt.alternate()` is true

prestwich: (as a drive-by while we're touching this code)

DaniPopes: I'll open a new PR for this

PR #450, opened 2023-12-12
feat: add `RevertReason` enum

### LINKED ISSUE #449: [Feature] Add RevertReason type
### Component

sol-types

currently

https://github.com/alloy-rs/core/blob/398d7e27b235f85f776c1b6b8fe2c9b081f2907a/crates/sol-types/src/types/error.rs#L411-L415

returns a String, which can be the Display implementation of:

https://github.com/alloy-rs/core/blob/398d7e27b235f85f776c1b6b8fe2c9b081f2907a/crates/sol-types/src/types/interface/mod.rs#L150
and `Revert` that prefixes the reason:

https://github.com/alloy-rs/core/blob/398d7e27b235f85f776c1b6b8fe2c9b081f2907a/crates/sol-types/src/types/error.rs#L103

so there's no easy way to get the actual `reason` that we need for RPC tracing for example.

proposing a new enum:

```rust
enum RevertReason {
  Error(ContractError),
  Raw(String)
}
``` 

### Additional context

_No response_

tcoratger:
## Motivation

Adding the `RevertReason` enum addresses the need to encapsulate and distinguish reasons for reverts.

Should close #449.

## Solution

Introducing the `RevertReason` enum offers a structured way to represent both detailed contract errors and raw string messages as reasons for reverts, enhancing clarity and flexibility in error handling within the codebase.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation

### DIFF
--- a/crates/sol-types/src/types/error.rs
+++ b/crates/sol-types/src/types/error.rs
@@ -1,12 +1,9 @@
 use crate::{
     abi::token::{PackedSeqToken, Token, TokenSeq, WordToken},
-    types::interface::GenericRevertReason,
-    GenericContractError, Result, SolInterface, SolType, Word,
+    types::interface::{GenericRevertReason, RevertReason},
+    Result, SolType, Word,
 };
-use alloc::{
-    string::{String, ToString},
-    vec::Vec,
-};
+use alloc::{string::String, vec::Vec};
 use alloy_primitives::U256;
 use core::{borrow::Borrow, fmt};
 
@@ -408,27 +405,20 @@ impl PanicKind {
 
 /// Decodes and retrieves the reason for a revert from the provided output data.
 ///
-/// Returns `None` if the content is not a valid ABI-encoded [`GenericContractError`] or
-/// a [UTF-8 string](String) (for Vyper reverts).
+/// This function attempts to decode the provided output data as a generic contract error
+/// or a UTF-8 string (for Vyper reverts) using the `RevertReason::decode` method.
+///
+/// If successful, it returns the decoded revert reason wrapped in an `Option`.
+///
+/// If both attempts fail, it returns `None`.
 pub fn decode_revert_reason(out: &[u8]) -> Option<GenericRevertReason> {
-    // Try to decode as a generic contract error.
-    if let Ok(error) = GenericContractError::abi_decode(out, true) {
-        return Some(error.into());
-    }
-
-    // If that fails, try to decode as a regular string.
-    if let Ok(decoded_string) = core::str::from_utf8(out) {
-        return Some(decoded_string.to_string().into());
-    }
-
-    // If both attempts fail, return None.
-    None
+    RevertReason::decode(out)
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::sol;
+    use crate::{sol, types::interface::SolInterface};
     use alloy_primitives::{address, hex, keccak256};
 
     #[test]
--- a/crates/sol-types/src/types/error.rs
+++ b/crates/sol-types/src/types/error.rs
@@ -1,5 +1,6 @@
 use crate::{
     abi::token::{PackedSeqToken, Token, TokenSeq, WordToken},
+    types::interface::GenericRevertReason,
     GenericContractError, Result, SolInterface, SolType, Word,
 };
 use alloc::{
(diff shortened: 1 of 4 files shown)


crates/sol-types/src/types/interface/mod.rs
```diff
@@ -369,6 +369,47 @@ impl<T> ContractError<T> {
...
+///
+/// This enum captures two possible scenarios for a revert:
+///
+/// - [`ContractError`](RevertReason::ContractError): Contains detailed error information, such as a
+///   specific [`Revert`] or [`Panic`] error.
+///
+/// - [`RawString`](RevertReason::RawString): Represents a raw string message as the reason for the
+///   revert.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum RevertReason<T> {
```
DaniPopes, line 385: Can you also implement `fmt::Display`?

crates/sol-types/src/types/interface/mod.rs
```diff
@@ -369,6 +369,47 @@ impl<T> ContractError<T> {
     }
 }
 
+/// Represents the reason for a revert in a generic contract error.
+pub(crate) type GenericRevertReason = RevertReason<Infallible>;
```
DaniPopes, line 373: This can be public too

DaniPopes (changes requested):
sg

crates/sol-types/src/types/interface/mod.rs
tcoratger: Done

crates/sol-types/src/types/interface/mod.rs
tcoratger: Done

crates/sol-types/src/types/error.rs
```diff
@@ -405,18 +406,19 @@ impl PanicKind {
...
 
-/// Returns the revert reason from the given output data. Returns `None` if the
-/// content is not a valid ABI-encoded [`GenericContractError`] or a [UTF-8
-/// string](String) (for Vyper reverts).
-pub fn decode_revert_reason(out: &[u8]) -> Option<String> {
+/// Decodes and retrieves the reason for a revert from the provided output data.
+///
+/// Returns `None` if the content is not a valid ABI-encoded [`GenericContractError`] or
+/// a [UTF-8 string](String) (for Vyper reverts).
+pub fn decode_revert_reason(out: &[u8]) -> Option<GenericRevertReason> {
```
mattsse, line 413: should we move this to the type instead:

```rust
impl<T: SolInterface> RevertReason<T> {
  
   fn decode() -> Option<Self>
}
```

tcoratger: @mattsse So you mean something like?

```rust
impl<T: SolInterface> RevertReason<T> {
    /// Decodes and retrieves the reason for a revert from the provided output data.
    ///
    /// Returns `None` if the content is not a valid ABI-encoded [`GenericContractError`] or
    /// a [UTF-8 string](String) (for Vyper reverts).
    pub fn decode(out: &[u8]) -> Option<Self> {
        // Try to decode as a generic contract error.
        if let Ok(error) = GenericContractError::abi_decode(out, true) {
            return Some(error.into());
        }

        // If that fails, try to decode as a regular string.
        if let Ok(decoded_string) = core::str::from_utf8(out) {
            return Some(decoded_string.to_string().into());
        }

        // If both attempts fail, return None.
        None
    }
}
```

And then?

```rust
/// Decodes and retrieves the reason for a revert from the provided output data.
///
/// Returns `None` if the content is not a valid ABI-encoded [`GenericContractError`] or
/// a [UTF-8 string](String) (for Vyper reverts).
pub fn decode_revert_reason(out: &[u8]) -> Option<GenericRevertReason> {
    RevertReason::decode(out)
}
```

DaniPopes: Yeah that makes sense

tcoratger: Changes made

crates/sol-types/src/types/interface/mod.rs
```diff
@@ -419,6 +419,34 @@ impl<T> From<String> for RevertReason<T> {
...
+    ///
+    /// This method attempts to decode the provided output data as a generic contract error
+    /// or a UTF-8 string (for Vyper reverts).
+    ///
+    /// If successful, it returns the decoded revert reason wrapped in an `Option`.
+    ///
+    /// If both attempts fail, it returns `None`.
+    pub fn decode(out: &[u8]) -> Option<Self> {
+        // Try to decode as a generic contract error.
+        if let Ok(error) = GenericContractError::abi_decode(out, true) {
```
mattsse, line 436: can we not use `ContractError::<T>::abi_decode` here?

tcoratger: yep true, change is made

DaniPopes (approved):
LGTM

PR #459, opened 2023-12-18
feat(primitives): add Signature type and utils

prestwich:
## Motivation

Add a primitive `Signature` type to support ongoing work in alloy transaction models

## Solution

- Add `Signature` to primitives crate
- Add `k256` feature
- Add `Address::from_public_key` convenience function

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/address.rs
+++ b/crates/primitives/src/bits/address.rs
@@ -426,9 +426,9 @@ impl Address {
         Address::from_slice(&digest[12..])
     }
 
-    #[cfg(feature = "k256")]
     /// Converts an ECDSA public key to its corresponding Ethereum address.
     #[inline]
+    #[cfg(feature = "k256")]
     pub fn from_public_key(pubkey: &k256::ecdsa::VerifyingKey) -> Self {
         use k256::elliptic_curve::sec1::ToEncodedPoint;
         let affine: &k256::AffinePoint = pubkey.as_ref();
--- a/crates/primitives/src/bits/address.rs
+++ b/crates/primitives/src/bits/address.rs
@@ -425,6 +425,16 @@ impl Address {
         let digest = keccak256(pubkey);
         Address::from_slice(&digest[12..])
     }
+
+    #[cfg(feature = "k256")]
+    /// Converts an ECDSA public key to its corresponding Ethereum address.
+    #[inline]
+    pub fn from_public_key(pubkey: &k256::ecdsa::VerifyingKey) -> Self {
+        use k256::elliptic_curve::sec1::ToEncodedPoint;
+        let affine: &k256::AffinePoint = pubkey.as_ref();
+        let encoded = affine.to_encoded_point(false);
+        Self::from_raw_public_key(&encoded.as_bytes()[1..])
+    }
 }
 
 #[cfg(test)]
--- a/crates/primitives/src/signature.rs
+++ b/crates/primitives/src/signature.rs
@@ -1,6 +1,5 @@
 use crate::{hex, ChainId, Uint, U256, U64};
 use alloc::vec::Vec;
-use alloy_rlp::length_of_length;
 use core::str::FromStr;
 
 /// Applies [EIP-155](https://eips.ethereum.org/EIPS/eip-155).
@@ -625,7 +624,7 @@ impl alloy_rlp::Encodable for crate::Signature {
 
     fn length(&self) -> usize {
         let payload_length = self.rlp_vrs_len();
-        payload_length + length_of_length(payload_length)
+        payload_length + alloy_rlp::length_of_length(payload_length)
     }
 }
 
--- a/crates/primitives/src/signature.rs
+++ b/crates/primitives/src/signature.rs
@@ -116,10 +116,10 @@ impl Parity {
     }
 
     /// Return the y-parity
-    pub fn y_parity(&self) -> bool {
+    pub const fn y_parity(&self) -> bool {
         match self {
-            Parity::V(v @ 0..=34) => v % 2 == 1,
-            Parity::V(v) => (v ^ 1) % 2 == 0,
+            Parity::V(v @ 0..=34) => *v % 2 == 1,
+            Parity::V(v) => (*v ^ 1) % 2 == 0,
             Parity::Parity(y) => *y,
         }
     }
@@ -139,10 +139,10 @@ impl Parity {
     }
 
     /// Apply EIP 155 to the V value. This is a nop for parity values.
-    pub const fn with_chain_id(&self, chain_id: ChainId) -> Self {
+    pub const fn with_chain_id(self, chain_id: ChainId) -> Self {
         let parity = match self {
-            Parity::V(v) => normalize_v_to_byte(*v) == 1,
-            Parity::Parity(y) => *y,
+            Parity::V(v) => normalize_v_to_byte(v) == 1,
+            Parity::Parity(_) => return self,
         };
 
         Self::V(to_eip155_v(parity as u8, chain_id))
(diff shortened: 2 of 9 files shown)


prestwich:
CI failure is a new clippy lint, unrelated to this work

crates/primitives/src/signature.rs
```diff
@@ -0,0 +1,855 @@
...
+        }
+    }
+
+    /// Return the y-parity as 0 or 1
+    pub fn y_parity_byte(&self) -> u8 {
+        self.y_parity() as u8
+    }
+
+    /// Invert the parity
+    pub fn inverted(&self) -> Self {
```
DaniPopes, line 133: fn invert(&mut self)?

crates/primitives/src/signature.rs
```diff
@@ -0,0 +1,855 @@
...
+            Parity::V(v) => v.encode(out),
+            Parity::Parity(false) => 27u8.encode(out),
+            Parity::Parity(true) => 28u8.encode(out),
+        }
+    }
+
+    fn length(&self) -> usize {
+        match self {
+            Parity::V(v) => v.length(),
+            Parity::Parity(_) => 1,
```
DaniPopes, line 180:
```suggestion
            Parity::Parity(v) => v.length(),
```

crates/primitives/src/signature.rs
```diff
@@ -0,0 +1,855 @@
...
+    pub fn from_rs_and_parity<T: Into<Parity>>(
+        r: U256,
+        s: U256,
+        parity: T,
+    ) -> Result<Self, SignatureError> {
+        Ok(Self { inner: (), v: parity.into(), r, s })
+    }
+}
+
+impl<S: Copy + Eq> Signature<S> {
```
DaniPopes, line 451: don't think this whole block needs Copy bound, it's ok since both generics we use do implement it tho

crates/primitives/src/signature.rs
```diff
@@ -0,0 +1,855 @@
...
+            27 => Self::Parity(false),
+            28 => Self::Parity(true),
+            v => Self::from(v),
+        })
+    }
+}
+
+/// An Ethereum ECDSA signature.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Signature<T: Copy + Eq> {
```
DaniPopes, line 199: Might need methods ported from here https://github.com/alloy-rs/alloy/pull/20/files#diff-097d91c911beebc907ad9aa358245953a05f69993d37ef026107af7c26d3c872

crates/primitives/src/signature.rs
DaniPopes, line 1: Can also add a macro for instantiating a signature from hex

DaniPopes (changes requested):
Mostly OK, might need to port from the other PR.

crates/primitives/src/signature.rs
prestwich: most of those exist but are renamed iirc

crates/primitives/src/signature.rs
prestwich: I've been avoiding mutating functions on signatures, as it becomes too easy to accidentally invalidate them. instead the API takes advantage of `Copy` a lot

crates/primitives/src/signature.rs
prestwich: thought about it, but given the instantiation can't be const and is always fallible, i think we can skip for now?

crates/primitives/src/signature/parity.rs
```diff
@@ -0,0 +1,198 @@
...
+            Parity::Eip155(v) => normalize_v_to_byte(v) == 1,
+            Parity::NonEip155(b) => b,
+            Parity::Parity(_) => return self,
+        };
+
+        Self::Eip155(to_eip155_v(parity as u8, chain_id))
+    }
+
+    #[cfg(feature = "k256")]
+    /// Determine the recovery id.
```
DaniPopes, line 117: Swap

crates/primitives/src/signature/sig.rs
```diff
@@ -0,0 +1,727 @@
...
+}
+
+impl From<&crate::Signature> for Vec<u8> {
+    #[inline]
+    fn from(value: &crate::Signature) -> Vec<u8> {
+        value.as_bytes().to_vec()
+    }
+}
+
+impl From<crate::Signature> for Vec<u8> {
```
DaniPopes, line 86: can these impls be generic over T?

crates/primitives/src/signature/sig.rs
```diff
@@ -0,0 +1,727 @@
...
+    }
+
+    #[cfg(feature = "rlp")]
+    /// Write the V to an RLP buffer without using EIP-155.
+    pub fn write_rlp_v(&self, out: &mut dyn alloy_rlp::BufMut) {
+        alloy_rlp::Encodable::encode(&self.v, out);
+    }
+
+    #[cfg(feature = "rlp")]
+    /// Write the VRS to the output. The V will always be 27 or 28.
```
DaniPopes, line 367: Swap doc and cfg

crates/primitives/src/signature/sig.rs
```diff
@@ -0,0 +1,727 @@
...
+    }
+
+    /// Modifies the recovery ID by dropping any [EIP-155] v value, converting
+    /// to a simple parity bool.
+    pub fn with_parity_bool(self) -> Self {
+        self.with_parity(self.v.to_parity_bool())
+    }
+}
+
+impl<S> Signature<S> {
```
DaniPopes, line 297: conversions for `Signature<_> -> Signature<k256>`?

DaniPopes (changes requested):

crates/primitives/src/signature/sig.rs
prestwich: currently the generic `Signature<T>` type is not reachable by downstream crates. only the aliased parameterized version is. so there's no purpose in making these generic, as they'd only ever be used with the alias

crates/primitives/src/signature/sig.rs
prestwich: again, because we're only exporting a parameterized version, we don't need any conversions. no program will be using multiple `Signature<T>` with different T

DaniPopes: What about the existing `<()>`?

prestwich: no downstream user can instantiate anything _other than_ `crate::Signature`

DaniPopes: Ah right though both () and k256 were exposed, but there's only one feature-gated export

crates/primitives/src/bits/address.rs
```diff
@@ -425,6 +425,16 @@ impl Address {
         let digest = keccak256(pubkey);
         Address::from_slice(&digest[12..])
     }
+
+    #[cfg(feature = "k256")]
```
DaniPopes, line 429: to the bottom

DaniPopes (approved):

crates/primitives/src/bits/address.rs
```diff
@@ -425,6 +425,16 @@ impl Address {
...
+    /// Converts an ECDSA public key to its corresponding Ethereum address.
+    #[inline]
+    pub fn from_public_key(pubkey: &k256::ecdsa::VerifyingKey) -> Self {
```
DaniPopes, line 432:
```suggestion
    /// Converts an ECDSA public key to its corresponding Ethereum address.
    #[inline]
    #[cfg(feature = "k256")]
    pub fn from_public_key(pubkey: &k256::ecdsa::VerifyingKey) -> Self {
```

DaniPopes (changes requested):

DaniPopes (approved):

PR #467, opened 2023-12-27
feature: sealed

prestwich:
Moves `Sealed` and `Sealable` into alloy-primitives

## Motivation

Commonly used across many potential networks and crates.

## Solution

Make `Sealed` and `Sealable` primitives

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -42,12 +42,6 @@ impl<T> Sealed<T> {
     pub const fn seal(&self) -> B256 {
         self.seal
     }
-
-    /// Geth the hash (alias for [`Self::seal`]).
-    #[inline(always)]
-    pub const fn hash(&self) -> B256 {
-        self.seal()
-    }
 }
 
 /// Sealeable objects.
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -53,11 +53,11 @@ impl<T> Sealed<T> {
 /// Sealeable objects.
 pub trait Sealable: Sized {
     /// Calculate the seal hash, this may be slow.
-    fn hash(&self) -> B256;
+    fn hash_slow(&self) -> B256;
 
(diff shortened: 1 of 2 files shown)


crates/primitives/src/sealed.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[inline(always)]
+    pub const fn hash(&self) -> B256 {
+        self.seal()
+    }
+}
+
+/// Sealeable objects.
+pub trait Sealable: Sized {
+    /// Calculate the seal hash, this may be slow.
+    fn hash(&self) -> B256;
```
mattsse, line 56: hmm, should this be hash_slow?

perhaps this could clash with `Sealed::hash`
`<Sealed Deref::Target>::hash`

mattsse (changes requested):
I think we also want `unseal(self) -> T`

https://github.com/paradigmxyz/reth/blob/22a450822ea808f22309b6e266af679b804fc13e/crates/primitives/src/header.rs#L526-L529

crates/primitives/src/sealed.rs
prestwich: easy 👍

mattsse (approved):

crates/primitives/src/sealed.rs
```diff
@@ -0,0 +1,82 @@
+use crate::B256;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+/// A consensus hashable item, with its memoized hash.
+///
+/// We do not implement
```
DaniPopes, line 6: ... ?

crates/primitives/src/sealed.rs
```diff
@@ -0,0 +1,82 @@
+use crate::B256;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
DaniPopes, line 3: attrs below docs

PR #471, opened 2023-12-29
chore: move define_udt! decl macro to sol! proc macro

### LINKED ISSUE #39: Move `define_udt!` decl macro to `sol!` proc macro
https://github.com/alloy-rs/core/blob/0c21fcc48eab0b52c20d3e98c739bac862aad8ff/crates/sol-types/src/types/udt.rs#L7

fgimenez:
## Motivation

Closes #39

## Solution

Follows the suggestion here https://github.com/alloy-rs/core/issues/39#issuecomment-1731942527, just moves the `define_udt!` content to proc macro in udt's `expand` function using `quote!`.

Besides moving the code I've used `expand_type(ty)` as `#underlying` and hardcoded `::alloy_sol_types::private::just_ok` instead of `$path`, let me know if that's ok. I've also added `#[automatically_derived]` annotations.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/sol-macro/src/expand/udt.rs
```diff
@@ -1,32 +1,120 @@
...
+
+            #[inline]
+            fn valid_token(token: &Self::Token<'_>) -> bool {
+                Self::type_check(token).is_ok()
+            }
+
+            #[inline]
+            fn type_check(token: &Self::Token<'_>) -> ::alloy_sol_types::Result<()> {
+                <#underlying as ::alloy_sol_types::SolType>::type_check(token)?;
+                ::alloy_sol_types::private::just_ok(token)
```
DaniPopes, line 93: If you don't mind, I think we can keep the user input with a `#[sol(type_check = "function_name")]` attribute. Otherwise I'll open an issue after.

fgimenez: ok, should be done here https://github.com/alloy-rs/core/pull/471/commits/e7e7b8f106e22d68d18ba56f9cd433404c6b410e PTAL

DaniPopes (approved):
Thanks!

PR #480, opened 2024-01-08
feat(json-abi): add full_signature

BrennerSpear:
## Motivation

I’m working on a ABI fragments database so we can fill in function and event data for contracts that haven’t been verified, but we having a matching ABI fragment for from another contract.

We’re going to have each row be unique based on the `full_signature`, which includes the name of each param, for events, whether the param is indexed or not, and for functions, the state mutability

In `ethers.js`, there is a fragment `FormatTypes.full` that can generate the full signature for you.

https://docs.ethers.org/v5/api/utils/abi/fragments/#fragments--output-formats

some maximally illustrative examples:
`function batchSwapExactIn(tuple(address pool, uint256 tokenInParam, uint256 tokenOutParam, uint256 maxPrice)[] swaps, address tokenIn, address tokenOut, uint256 totalAmountIn, uint256 minTotalAmountOut) payable returns (uint256 totalAmountOut)`

`event SetupDirectDebit(address indexed debtor, address indexed receiver, tuple(uint256 amount, uint256 startTime, uint256 interval) info)`

## Solution

I've added a `.full_signature()` function to the `Function` and `Event` structs

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] No Breaking changes

## Notes

Functions already have `signature_full` that includes the output types. This is pretty close to `full_signature` as a name. Are we fine with that or do we need to think of a different name for `full_signature`?

I thought about trying to overload some of the functions instead of making additional functions prepended with `full_`, but I don’t think it’d make it any cleaner.

### DIFF
--- a/crates/json-abi/src/item.rs
+++ b/crates/json-abi/src/item.rs
@@ -503,7 +503,7 @@ impl Function {
     }
 
     /// Returns this function's full signature including names of params:
-    /// `event $name($($inputs names),*) state_mutability returns ($(outputs names),*)`.
+    /// `function $name($($inputs $names),*) state_mutability returns ($($outputs $names),*)`.
     ///
     /// This is a full human-readable string, including all parameter names, any optional modifiers
     /// (e.g. indexed, public, etc) and white-space to aid in human readability. This is useful for
--- a/crates/json-abi/src/item.rs
+++ b/crates/json-abi/src/item.rs
@@ -498,7 +498,7 @@ impl Function {
     /// This is the same as [`signature`](Self::signature), but also includes
     /// the output types.
     #[inline]
-    pub fn signature_full(&self) -> String {
+    pub fn signature_with_outputs(&self) -> String {
         signature(&self.name, &self.inputs, Some(&self.outputs))
     }
 
@@ -506,7 +506,7 @@ impl Function {
     /// `function $name($($inputs $names),*) state_mutability returns ($($outputs $names),*)`.
     ///
     /// This is a full human-readable string, including all parameter names, any optional modifiers
-    /// (e.g. indexed, public, etc) and white-space to aid in human readability. This is useful for
+    /// (e.g. view, payable, pure) and white-space to aid in human readability. This is useful for
     /// storing a string which can still fully reconstruct the original Fragment
     #[inline]
     pub fn full_signature(&self) -> String {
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):
LGTM. Thanks!

crates/json-abi/src/item.rs
```diff
@@ -502,6 +502,17 @@ impl Function {
...
     }
 
+    /// Returns this function's full signature including names of params:
+    /// `event $name($($inputs names),*) state_mutability returns ($(outputs names),*)`.
+    ///
+    /// This is a full human-readable string, including all parameter names, any optional modifiers
+    /// (e.g. indexed, public, etc) and white-space to aid in human readability. This is useful for
+    /// storing a string which can still fully reconstruct the original Fragment
+    #[inline]
+    pub fn full_signature(&self) -> String {
```
DaniPopes, line 512: I think we should rename the previous `signature_full` to `signature_with_outputs`

crates/json-abi/src/item.rs
```diff
@@ -563,6 +574,12 @@ impl Event {
         event_signature(&self.name, &self.inputs)
     }
 
+    /// Returns this event's full signature: `event $name($($inputs indexed names),*)`.
```
DaniPopes, line 577: Can you also add the paragraph from Function's `full_signature`?

crates/json-abi/src/item.rs
BrennerSpear: I like this idea too. to remove `signature_full` would be a breaking change. Maybe we just add the `signature_with_outputs` as an alias and warn `signature_full` is going to be removed?

(not sure how y'all do SemVer when you're not at 1.0 yet. I guess removing it could come when it becomes 1.0.0!)

crates/json-abi/src/item.rs
BrennerSpear: will do

crates/json-abi/src/item.rs
DaniPopes: That's OK, the next release we'll be doing soon is already breaking

BrennerSpear: updated to `signature_with_outputs`

DaniPopes (approved):

yorhodes:
is this relevant? https://www.4byte.directory/

PR #489, opened 2024-01-20
Create SECURITY.md

paulmillr:
It's important for reporting.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


paulmillr:
Why close?

It’s impossible to report vulnerabilities thru github private reporting now. You also don’t have an email specified. 

Folks will just drop 0-days if you keep it like that.

DaniPopes:
I don't see how a default SECURITY.md template is better than nothing. But you're right, we should have one: https://github.com/alloy-rs/core/pull/494

Issue #490, opened 2024-01-20
[Bug] Recursive ABI parsing causes DoS


paulmillr:
### Component

dyn-abi

### Operating System

None

### Describe the bug

More details in https://github.com/paulmillr/micro-eth-signer/discussions/20.

https://github.com/alloy-rs/core/issues/392 and https://github.com/alloy-rs/core/pull/454 are slightly related

PoC:

```rs
use alloy_dyn_abi::{DynSolType, DynSolValue};
fn main() {
[...]

PR #491, opened 2024-01-20
test: dyn decode alloc test

mattsse:
some debugging for #490

I believe @DaniPopes already identified the alloc issue here


DaniPopes:
https://github.com/alloy-rs/core/pull/495

Issue #482, opened 2024-01-10
generate bindings for constructors


rouzwelt:
### Component

sol-types, sol! macro

### Describe the feature you would like

right now when using `sol!` macro for some json artifact or inline solidity contract, all the types for functions/events/errors are generated and can be easily instantiated and Abi-encoded/decoded, as they would appear in `contractNameCalls`, `contractNameEvents` and `contractNameErrors`, however there is no similar thing for constructor of the underlying contract.

So Im suggesting that a similar functionality be provided for constructors, possibly with an option of building the deploy data with the contract `bytecode` if it was available.

DaniPopes:
https://github.com/alloy-rs/core/pull/493

PR #493, opened 2024-01-22
feat: add constructorCall to `sol!`

leruaa:
## Motivation

I needed to encode constructor args to create a contract creation tx, and saw #482.

### In `sol-types`

* Adding `SolConstructor` trait 
* Handle constructor in `function::expand()` (creating a constructorCall struct  that impl `SolConstructor`)

### In `json-abi`

* impl `ToSol` for `Constructor`
* Adding `AbiFunctionKw::Constructor` variant
* Update fixtures in "tests/abi"
 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-abi/src/to_sol.rs
+++ b/crates/json-abi/src/to_sol.rs
@@ -67,9 +67,9 @@ impl ToSol for JsonAbi {
         let mut its = InternalTypes::new();
         its.visit_abi(self);
         fmt!(its.0);
-        fmt!(self.constructor());
         fmt!(self.errors());
         fmt!(self.events());
+        fmt!(self.constructor());
         fmt!(self.fallback);
         fmt!(self.receive);
         fmt!(self.functions());
--- a/crates/json-abi/src/to_sol.rs
+++ b/crates/json-abi/src/to_sol.rs
@@ -1,5 +1,5 @@
 use crate::{
-    item::{Error, Event, Fallback, Function, Receive},
+    item::{Constructor, Error, Event, Fallback, Function, Receive},
     EventParam, InternalType, JsonAbi, Param, StateMutability,
 };
 use alloc::{collections::BTreeSet, string::String, vec::Vec};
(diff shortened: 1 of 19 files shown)


crates/json-abi/src/to_sol.rs
```diff
@@ -67,6 +67,7 @@ impl ToSol for JsonAbi {
         let mut its = InternalTypes::new();
         its.visit_abi(self);
         fmt!(its.0);
+        fmt!(self.constructor());
```
DaniPopes, line 70: I think this should be right after events.

crates/sol-macro/src/expand/function.rs
```diff
@@ -24,10 +24,17 @@ use syn::Result;
...
+    let ItemFunction { attrs, parameters, returns, name, kind, .. } = function;
+
+    if matches!(kind, FunctionKind::Constructor(_)) {
+        return expand_constructor(cx, function);
+    }
+
+    let Option::Some(_) = name else {
+        // ignore functions without names (modifiers...)
         return Ok(quote!());
     };
```
DaniPopes, line 36:
```suggestion
    if name.is_none() {
        // ignore functions without names (modifiers...)
        return Ok(quote!());
    }
```

crates/sol-types/src/types/function.rs
```diff
@@ -101,3 +101,41 @@ pub trait SolCall: Sized {
...
+        (self.tokenize().total_words() * Word::len_bytes()).saturating_sub(offset)
+    }
+
+    /// ABI encode the call to the given buffer.
+    #[inline]
+    fn abi_encode(&self) -> Vec<u8> {
+        let mut out = Vec::with_capacity(self.abi_encoded_size());
+        out.reserve(self.abi_encoded_size());
+        out.extend(crate::abi::encode_sequence(&self.tokenize()));
+        out
```
DaniPopes, line 139:
```suggestion
        crate::abi::encode_sequence(&self.tokenize())
```

DaniPopes (changes requested):

DaniPopes (approved):
LGTM, thanks!

PR #495, opened 2024-01-23
wip: recursion mitigations

prestwich:
## Motivation

discussion aid for ongoing work on #490 

## Solution

- recursion limit of 16
- check that the decoder contains sufficient words before allocating memory

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes (additional enum variants)


crates/sol-types/src/abi/decoder.rs
```diff
@@ -85,6 +90,16 @@ impl<'de> Decoder<'de> {
         self.buf.len().checked_sub(self.offset)
     }
 
+    /// Returns the number of words in the remaining buffer.
+    #[inline]
+    pub const fn remaining_words(&self) -> Option<usize> {
+        if let Some(remaining) = self.remaining() {
```
prestwich, line 96: should be self.remaining.map

crates/dyn-abi/src/token.rs
```diff
@@ -153,6 +153,10 @@ impl<'a> DynToken<'a> {
                 // `enc(X)`. But known-good test vectors have it relative to the
                 // word AFTER the array size
                 let mut child = child.raw_child();
+                let rem = child.remaining_words();
+                if rem.is_none() || rem.expect("checked") < size {
```
prestwich, line 157: should be ok_or with ? operator

crates/sol-types/src/abi/decoder.rs
prestwich: map is not `const`

prestwich:
@DaniPopes I'm unsure the best way to fix the proptest

DaniPopes:
> @DaniPopes I'm unsure the best way to fix the proptest

Fixed in 0e9edd9f0426ab2d72e9435523ef53eb8192c5a9

DaniPopes:
@prestwich anything else to be done here? Looks good enough to merge already

prestwich:
Are we confident this resolves the issues? or do we think additional work is needed?

DaniPopes:
I'd say so, in any case I would like to get this in for the next release.

DaniPopes (approved):

PR #499, opened 2024-01-24
feat(`primitives`): Add `y_parity_byte_non_eip155` to `Parity`

Evalir:
## Motivation

In the case of a signature without EIP155 applied, the `y_parity` byte should be returned as 27 or 28. `y_parity_byte` only returns it as `{0,1}`.

## Solution

 This adds a fn `y_parity_byte_non_eip155` which, when not having an EIP155 parity, we get 27 or 28, or `None` otherwise.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/parity.rs
+++ b/crates/primitives/src/signature/parity.rs
@@ -83,6 +83,15 @@ impl Parity {
         self.y_parity() as u8
     }
 
+    /// Return the y-parity byte as 27 or 28,
+    /// In the case of a non-EIP155 signature.
+    pub const fn y_parity_byte_legacy(&self) -> Option<u8> {
+        match self {
+            Self::NonEip155(v) => Some(*v as u8 + 27),
+            _ => None,
+        }
+    }
+
     /// Inverts the parity.
     pub const fn inverted(&self) -> Self {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
defer to @prestwich here

onbjerg (approved):
lgtm, any follow up PR needed here to solve the bug in anvil?

crates/primitives/src/signature/parity.rs
```diff
@@ -83,6 +83,15 @@ impl Parity {
         self.y_parity() as u8
     }
 
+    /// Return the y-parity byte as 27 or 28,
+    /// In the case of a non-EIP155 signature.
```
onbjerg, line 87: nit 

```suggestion
    /// Return the y-parity byte as 27 or 28,
    /// in the case of a non-EIP155 signature.
```

Evalir:
nope @onbjerg, but we'll need a minor release of core to bump on `alloy-rs/alloy`.

onbjerg:
lgtm

gakonst (approved):

PR #508, opened 2024-01-28
perf(sol-macro): use a lookup table when generating `SolInterface::abi_decode_raw`

DaniPopes:
Faster and maybe less binary size


mattsse (approved):

DaniPopes:
Tested locally against foundry cheatcodes

PR #509, opened 2024-02-04
fix: Signature bincode serialization

Wollac:
## Motivation

The current way the serialization/deserialization of `Signature` is implemented only works for human readable (JSON) formats. However, binary formats like `bincode` that do not support `Deserializer::deserialize_identifier` are not supported.

## Solution

Similar to #221, this PR uses the `is_human_readable()` flag to distinguish between using the current serialization (`SerializeMap`) or a `bincode` compatible version.
For this, the serialization is implemented as a tuple `(r: U256, s: U256, v: u64)`.

### DIFF
--- a/crates/primitives/src/signature/parity.rs
+++ b/crates/primitives/src/signature/parity.rs
@@ -92,6 +92,15 @@ impl Parity {
         }
     }
 
+    /// Return the corresponding u64 V value.
+    pub const fn to_u64(&self) -> u64 {
+        match self {
+            Self::Eip155(v) => *v,
+            Self::NonEip155(b) => *b as u64 + 27,
+            Self::Parity(b) => *b as u64,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


crates/primitives/src/signature/parity.rs
```diff
@@ -92,6 +92,15 @@ impl Parity {
         }
     }
 
+    /// Return the corresponding u64 V value.
+    pub const fn to_u64(&self) -> u64 {
```
prestwich, line 96: should be a `From/Into` implementation?

Wollac: Could be, yes. To me it felt more like the `y_parity_byte` which is not a `From<u8>` either.
But I don't have a strong opinion at all, I can change it to a `From`. Whatever you prefer.

DaniPopes (approved):
lgtm

PR #510, opened 2024-02-04
feat(sol-macro): add `#[sol(rpc)]` to generate type-safe provider contract calls

DaniPopes:
Adds `#[sol(rpc)]` to generate type-safe bindings for creating JSON-RPC `eth_call`s to on-chain contracts (alloy-contract, alloy-providers)

See https://github.com/alloy-rs/alloy/pull/182 for more details, examples and tests


crates/sol-macro/src/expand/contract.rs
```diff
@@ -200,6 +204,137 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
+                        ::alloy_contract::DeploymentCallBuilder::new_deployment(provider, data)
+                    }
+                },
+            )
+        }).unzip();
+        quote! {
+            #[doc = #new_fn_doc]
+            #[inline]
+            pub const fn new<P: ::alloy_contract::private::Provider>(
+                address: ::alloy_sol_types::private::Address,
```
gakonst, line 261: Does `private` work when you import from a 3rd party package?

DaniPopes: yes, it's `pub mod private` :smile:

crates/sol-macro/src/expand/contract.rs
```diff
@@ -200,6 +204,137 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
+            "A [`{contract_name}`](self) instance.\n\
+             \n\
+             Contains type-safe methods for interacting with an on-chain instance of the\n\
+             [`{contract_name}`](self) contract located at a given `address`, using a given\n\
+             provider `P`.\n\
+             \n\
+             See the [module-level documentation](self) for all the available methods."
+        );
+        let (deploy_fn, deploy_method) = bytecode.is_some().then(|| {
+            let doc_str = "todo";
```
onbjerg, line 227: to be addressed in this PR or a follow up?

DaniPopes: Here

onbjerg (approved):
lgtm

Issue #161, opened 2023-06-30
[Bug] Add tests to verify ABI `encode_packed` correctness


DaniPopes:
### Component

sol-type, dyn-abi

### Operating System

Linux

### Describe the bug

AFAICT there are no `encode_packed` tests

PR #516, opened 2024-02-07
fix(dyn-abi): abi_encode_packed attempt to subtract with overflow

bernard-wagner:
Fixes 515. Hoping to expand tests so that #161 is also addressed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,5 +69,5 @@ proptest-derive = "0.4"
 rand = { version = "0.8", default-features = false }
 ruint = { version = "1.11.1", default-features = false, features = ["alloc"] }
 ruint-macro = { version = "1", default-features = false }
-winnow = { version = "0.5.19", default-features = false, features = ["alloc"] }
+winnow = { version = "0.5.39", default-features = false, features = ["alloc"] }
 postgres-types = "0.2.6"
--- a/crates/dyn-abi/src/value.rs
+++ b/crates/dyn-abi/src/value.rs
@@ -787,3 +787,83 @@ impl DynSolValue {
         self.as_fixed_seq().map(Self::encode_seq)
     }
 }
+
+
+#[cfg(test)]
+mod tests {
+    use alloy_primitives::U256;
+    use crate::DynSolValue;
+
+    #[test]
+    fn abi_encode_packed_uint32() {
+        // abi.encodePacked(uint32(42))
+        let byte_size = 4;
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -69,5 +69,5 @@ proptest-derive = "0.4"
 rand = { version = "0.8", default-features = false }
 ruint = { version = "1.11.1", default-features = false, features = ["alloc"] }
 ruint-macro = { version = "1", default-features = false }
-winnow = { version = "0.5.19", default-features = false, features = ["alloc"] }
+winnow = { version = "0.5.39", default-features = false, features = ["alloc"] }
```
bernard-wagner, line 72: @DaniPopes I needed to change this to get my local build working again after https://github.com/alloy-rs/core/pull/507

Not sure why CI checkout the correct version.

crates/dyn-abi/src/value.rs
```diff
@@ -785,3 +787,83 @@ impl DynSolValue {
...
+            DynSolValue::String("test".to_owned()),
+            DynSolValue::Bytes(vec![0xde, 0xad]),
+        ];
+        let encoded = values.iter().map(|v| v.abi_encode_packed()).flatten().collect::<Vec<_>>();
+        assert_eq!(encoded, [255, b't', b'e', b's', b't', 0xde, 0xad]);
+    }
+
+    #[test]
+    fn abi_encode_packed_array() {
+        // abi.encodePacked([[uint32(1),uint32(2)], [uint32(3), uint32(4)]])
```
bernard-wagner, line 847: This is failing at the moment and will need more work, keeping PR in WIP for now.

Cargo.toml
DaniPopes: Because the lockfile is not checked in, you have to run cargo update locally. But yes this should've been bumped

DaniPopes (commented):
Sorry I just saw this PR after opening https://github.com/alloy-rs/core/pull/517. I will add your testcases to that PR and credit you in the commits.
Done in https://github.com/alloy-rs/core/pull/517/commits/08a702a50ac631ce88751baf2f97b7f6e9cb4ff5

Cargo.toml
DaniPopes: can you open a separate PR for this?

crates/dyn-abi/src/value.rs
DaniPopes: this is because `uint32` is encoded as 4 bytes while you're testing against `U256::to_be_bytes()` which encodes it as 32.

PR #519, opened 2024-02-07
fix(dyn-abi): abi-encode-packed always pads arrays

bernard-wagner:
## Motivation

`abi.encodePacked()` always pads array elements as 32-bytes.

The behavior for tuples and structs is worth discussing since the solidity encoder does not support either. Tuples are useful at the top-level, but should probably be disallowed when nested.

Compare `fixed_array_of_addresses` to the equivalent Solidity code:
```
address[2] memory fixed_array_of_addresses;
fixed_array_of_addresses[0] = 0x1111111111111111111111111111111111111111;
fixed_array_of_addresses[1] = 0x2222222222222222222222222222222222222222;
abi.encodePacked(fixed_array_of_addresses); // 0x00000000000000000000000011111111111111111111111111111111111111110000000000000000000000002222222222222222222222222222222222222222
```


bernard-wagner:
@DaniPopes hoping to get your thoughts on this.

DaniPopes:
I see, this makes sense:
> The encoding of an array is the concatenation of the encoding of its elements with padding.

We should probably do the same as Solidity, which disables all structs and nested dynamic arrays: https://github.com/ethereum/solidity/blob/bf59ff8bdf4fe104e558cb697abfc8f3b33daaba/libsolidity/analysis/TypeChecker.cpp#L59
We can do this in a follow-up PR.

Issue #525, opened 2024-02-12
[Bug] Parsing abi file without name field fails


ozgrakkurt:
### Component

json-abi

### What version of Alloy are you on?

0.6.x

### Operating System

macOS (Apple Silicon)

### Describe the bug

Parsing the given abi fails with: `missing field "name"` error

```json
[...]

PR #526, opened 2024-02-12
fix(json-abi): accept nameless `Param`s

DaniPopes:
Fixes https://github.com/alloy-rs/core/issues/525


ozgrakkurt:
@DaniPopes thanks!!!!!!

PR #527, opened 2024-02-12
feat(sol-macro): provide a way to override import paths for dependencies

DaniPopes:
Allows creating `sol!` wrappers to specify the import path for dependencies so downstream users don't have to worry about the dependency import problem.

For example `alloy_core::sol!` is a wrapper that provides `#![sol(alloy_sol_types = $crate::sol_types)]`.


crates/core/src/lib.rs
```diff
@@ -32,10 +32,25 @@ pub use alloy_json_abi as json_abi;
...
+pub use sol_types::sol;
 
 #[cfg(feature = "rlp")]
 #[doc(inline)]
 pub use alloy_rlp as rlp;
+
+/// [`sol!`](alloy_sol_types::sol!) macro wrapper to route imports to the correct crate.
+///
+/// See [`sol!`](alloy_sol_types::sol!) for the actual macro documentation.
+#[cfg(all(not(doc), feature = "sol-types"))]
```
DaniPopes, line 46: `not(doc)` so that generated documentation (e.g. docs.rs) displays the `sol!` re-export instead of this wrapper.

mattsse (approved):

PR #531, opened 2024-02-16
feat: derive Allocative on FixedBytes

Rjected:
## Motivation

Recently I was introduced to https://github.com/facebookexperimental/allocative, which seems promising as an easy way to collect granular information about certain data structures' memory usage in reth.

## Solution

Introduce the `allocative` feature into `primitives`, and to start, derive the `Allocative` trait on FixedBytes.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,3 +71,4 @@ ruint = { version = "1.11.1", default-features = false, features = ["alloc"] }
 ruint-macro = { version = "1", default-features = false }
 winnow = { version = "0.5.39", default-features = false, features = ["alloc"] }
 postgres-types = "0.2.6"
+allocative = { version = "0.3.2", default-features = false }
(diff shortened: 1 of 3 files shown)


onbjerg (approved):
imo ok since its cfg gated

DaniPopes:
Fails on 1.65 because it has higher MSRV

DaniPopes:
@Rjected do you want to add it to more types?

Rjected:
> @Rjected do you want to add it to more types?

I do, as many as possible

gakonst:
Yep this will improve profiling on Reth

Rjected:
@DaniPopes I think this is good for now, will open another PR when I get around to bytes stuff

.github/workflows/ci.yml
```diff
@@ -16,8 +16,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ["stable", "beta", "nightly", "1.65"] # MSRV
+        rust: ["stable", "beta", "nightly"]
```
prestwich, line 19: are we updating our MSRV policy?

DaniPopes: No, just not testing `--all-features` because this feature requires a higher MSRV

DaniPopes (approved):

PR #545, opened 2024-02-29
Make `Bytes` map to `Bytes` in `SolType`

rachel-bousfield:
## Motivation

`alloy-sol-types` exposes a `sol_data::T` variant for each of the primitive types. For example, `alloy-primitives::FixedBytes` is represented by an `alloy-sol-types::FixedBytes`, whose's `SolType` impl points back to the former via the associated `RustType`.
```rs
impl<const N: usize> SolType for FixedBytes<N>
where
    ByteCount<N>: SupportedFixedBytes,
{
    type RustType = RustFixedBytes<N>;   // alias for the primitive
    // rest of impl   
}
```

In each case `RustType` points back to the alloy primitive, except for `Bytes`. Currently, `Bytes` instead maps to `Vec<u8>`. 
```rs
impl SolType for Bytes {
    type RustType = Vec<u8>;
}
```

This creates issues for encoding & decoding patterns in SDKs like Stylus, which declare traits like.
```rs
pub trait AbiType {
    type SolType: SolType<RustType = Self>;
}
```

In particular, we lose the bidirectional relationship the other primitives have. The current work-around in Stylus is to declare a second, incompatible `Bytes` type that implements `SolType` with a corresponding`SolValueType`, but this is suboptimal since it's both confusing and not officially supported by Alloy.
```rs
/// **Note:** this trait is an implementation detail. As such, it should not
/// be implemented directly unless implementing a custom
/// [`SolType`](crate::SolType), which is also discouraged. Consider
/// using [`SolValue`](crate::SolValue) instead.
ρ trait SolTypeValue<T: super―SolType> { ... }
```

## Solution

The Solution is to make `Bytes` map to `Bytes`, which it turns out is a very small diff.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] breaking changes


crates/sol-macro/src/expand/ty.rs
```diff
@@ -98,7 +98,7 @@ pub(crate) fn rec_expand_rust_type(ty: &Type, crates: &ExternCrates, tokens: &mu
         Type::Address(span, _) => quote_spanned! {span=> #alloy_sol_types::private::Address },
         Type::Bool(span) => return Ident::new("bool", span).to_tokens(tokens),
         Type::String(span) => quote_spanned! {span=> #alloy_sol_types::private::String },
-        Type::Bytes(span) => quote_spanned! {span=> #alloy_sol_types::private::Vec<u8> },
+        Type::Bytes(span) => quote_spanned! {span=> #alloy_sol_types::private::Bytes },
```
mattsse, line 101: I think this should work, but not entirely sure @DaniPopes 

I was wondering if there's any ambiguity wrt byte[] vs bytes, we previously had this issue in abigen!

rachel-bousfield:
@prestwich Regarding the first point, there's currently a `macro_rules` `bytes!` macro that lets you create a `Bytes` from string literals. If we were to move to a proc macro, `bytes![...]` could expand to `Bytes::from(vec![...])` for non-str inputs.
```rs
macro_rules! bytes {
    () → {
        $crate::Bytes::new()
    };

    ($($s:literal)+) → {{
        // force const eval
        const STATIC_BYTES: &'static [u8] = &$crate::hex!($($s)+);
        $crate::Bytes::from_static(STATIC_BYTES)
    }};
}
```

For the second point, something that's nice about this is that `bytes` now expands to `Bytes`. You'd need to put `[]uint8` in an event, say, for the generated `struct` to actually have a `Vec<u8>` if I understand things correctly.

prestwich:
you can add a macro rules arm like this:
```
    [$($inner:literal),+] => {{
        // force const eval
        const STATIC_BYTES: &'static [u8] = &[$($inner),+];
        $crate::Bytes::from_static(STATIC_BYTES)
    }}
```

crates/sol-macro/src/expand/ty.rs
prestwich: sol_types does not have the ambiguity in usage that abigen had because it always fully specifies the solidity type via the type system. 

the goal of this PR is to remove ambiguity in the _mapping between_ rust and sol_types by establishing that `Vec<u8>` corresponds to `uint8[]` and `Bytes` corresponds to `bytes`. This allows stylus to assume the mapping is bijective and remove some user-facing complexity

DaniPopes (approved):

PR #551, opened 2024-03-01
Prestwich/dyn sol error

prestwich:
## Motivation

Work on bringing dyn-abi up to feature parity with sol-types

## Solution

Add `DynSolError` to match `DynSolType` and `DynSolEvent`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/dyn-abi/src/dynamic/error.rs
```diff
@@ -0,0 +1,112 @@
+use crate::{DynSolType, DynSolValue, Error, Result};
+use alloc::vec::Vec;
+use alloy_primitives::{keccak256, Selector};
+
+const PANIC_SELECTOR: Selector = Selector::new([0x4e, 0x48, 0x7b, 0x71]);
+const REVERT_SELECTOR: Selector = Selector::new([0x08, 0xc3, 0x79, 0xa0]);
```
Evalir, line 6: Can we add some doc comments with the string version of these? (`Panic(uint256)` & `Error(string)`)

DaniPopes: Could also be `alloy_sol_types::Revert::SELECTOR`

crates/dyn-abi/src/dynamic/error.rs
```diff
@@ -0,0 +1,112 @@
...
+/// A dynamic ABI error.
+///
+/// This is a representation of a Solidity error, which can be used to decode
+/// error events.
+#[derive(Debug, Clone, PartialEq)]
+pub struct DynSolError {
+    /// Error selector.
+    pub(crate) selector: Selector,
+    /// Error body types.
+    pub(crate) body: DynSolType,
```
DaniPopes, line 17: `Vec<DynSolType>`?

crates/dyn-abi/src/dynamic/error.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub const fn new(selector: Selector, body: DynSolType) -> Self {
+        Self { selector, body }
+    }
+
+    /// Error selector is the first 4 bytes of the keccak256 hash of the error
+    /// declaration.
+    pub const fn selector(&self) -> Selector {
+        self.selector
+    }
+
```
DaniPopes, line 56: `body` getter

DaniPopes (changes requested):

crates/dyn-abi/src/dynamic/error.rs
prestwich: `DynSolEvent` uses `DynSolType` for body. could change both?

PR #553, opened 2024-03-04
feat: rlp encoding for logs with generic event data

### LINKED ISSUE #544: [Bug] `Log` doesn't allow implementation of traits for non-default generic parameter
### Component

primitives

0.6.3

### Operating System

Linux

### Describe the bug

The `Log` generic uses a default type `= LogData`. This means that for any traits that are implemented for `Log`, the traits are only implemented for `Log::<LogData>`. Given that Rust does not allow implementation of external traits on external types, this prohibits people from implementing traits on the `Log` type when using a locally defined `CustomLogData`.

E.g. when trying to `impl alloy_rlp::Encodable for LogData<CustomLogData>` this would fail with the compiler error:

```bash
only traits defined in the current crate can be implemented for types defined outside of the crate define and implement a trait or new type instead
```

To solve this, the trait implementations in https://github.com/alloy-rs/core/blob/main/crates/primitives/src/log.rs#L120 should be generic; e.g. `impl<LogDataT: SomeTrait> alloy_rlp::Encodable for Log<LogDataT> { .. }`.

prestwich:
closes #544

## Motivation

Allow rlp encoding of logs after their data has been ABI decoded.

## Solution

Add a set of methods and a blanket Encode implementation for 
`Log<T> where for<'a> &'a T: Into<LogData>,`

Note that this change does not allow decoding, as that would require embedding the ABI error into the RLP error decode impl

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/log.rs
+++ b/crates/primitives/src/log.rs
@@ -122,21 +122,49 @@ impl Log {
     }
 }
 
+impl<T> Log<T>
+where
+    for<'a> &'a T: Into<LogData>,
+{
+    /// Creates a new log.
+    #[inline]
+    pub const fn new_from_event_unchecked(address: Address, data: T) -> Self {
+        Self { address, data }
+    }
+
+    /// Creates a new log from an deserialized event.
+    pub fn new_from_event(address: Address, data: T) -> Option<Self> {
+        let this = Self::new_from_event_unchecked(address, data);
+        (&this.data).into().is_valid().then_some(this)
+    }
+
+    /// Reserialize the data.
+    #[inline]
+    pub fn reserialze(&self) -> Log<LogData> {
+        Log { address: self.address, data: (&self.data).into() }
+    }
+}
+
 #[cfg(feature = "rlp")]
-impl alloy_rlp::Encodable for Log {
+impl<T> alloy_rlp::Encodable for Log<T>
+where
+    for<'a> &'a T: Into<LogData>,
+{
     fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
+        let this = self.reserialze();
         let payload_length =
-            self.address.length() + self.data.data.length() + self.data.topics.length();
+            this.address.length() + this.data.data.length() + this.data.topics.length();
 
         alloy_rlp::Header { list: true, payload_length }.encode(out);
-        self.address.encode(out);
-        self.data.topics.encode(out);
-        self.data.data.encode(out);
+        this.address.encode(out);
+        this.data.topics.encode(out);
+        this.data.data.encode(out);
     }
 
     fn length(&self) -> usize {
+        let this = self.reserialze();
         let payload_length =
-            self.address.length() + self.data.data.length() + self.data.topics.length();
+            this.address.length() + this.data.data.length() + this.data.topics.length();
         payload_length + alloy_rlp::length_of_length(payload_length)
     }
 }
--- a/crates/primitives/src/log.rs
+++ b/crates/primitives/src/log.rs
@@ -140,7 +140,7 @@ where
 
     /// Reserialize the data.
     #[inline]
-    pub fn reserialze(&self) -> Log<LogData> {
+    pub fn reserialize(&self) -> Log<LogData> {
         Log { address: self.address, data: (&self.data).into() }
     }
 }
@@ -151,7 +151,7 @@ where
     for<'a> &'a T: Into<LogData>,
 {
     fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
-        let this = self.reserialze();
+        let this = self.reserialize();
         let payload_length =
             this.address.length() + this.data.data.length() + this.data.topics.length();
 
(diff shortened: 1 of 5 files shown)


gakonst:
@Rjected ptal

crates/sol-types/tests/doctests/events.rs
```diff
@@ -61,6 +62,28 @@ fn event() {
...
+        a: [0x11; 32].into(),
+        b: U256::from(1u64),
+        c: keccak256("Hello World"),
+        d: Vec::new(),
+    };
+
+    let rlpable_log = Log::<MyEvent>::new_from_event_unchecked(Default::default(), event);
+
+    let mut rlp_encoded = vec![];
+    rlpable_log.encode(&mut rlp_encoded);
```
Rjected, line 77: should add an assert that `.length()` == `rlp_encoded.len()`

crates/primitives/src/log.rs
```diff
@@ -116,21 +116,49 @@ impl Log {
...
+
+    /// Creates a new log from an deserialized event.
+    pub fn new_from_event(address: Address, data: T) -> Option<Self> {
+        let this = Self::new_from_event_unchecked(address, data);
+        (&this.data).into().is_valid().then_some(this)
+    }
+
+    /// Reserialize the data.
+    #[inline]
+    pub fn reserialze(&self) -> Log<LogData> {
```
Rjected, line 137:
```suggestion
    pub fn reserialize(&self) -> Log<LogData> {
```

crates/primitives/src/log.rs
```diff
@@ -116,21 +116,49 @@ impl Log {
...
+        Log { address: self.address, data: (&self.data).into() }
+    }
+}
+
 #[cfg(feature = "rlp")]
-impl alloy_rlp::Encodable for Log {
+impl<T> alloy_rlp::Encodable for Log<T>
+where
+    for<'a> &'a T: Into<LogData>,
+{
```
Rjected, line 146: ah I see, can just be converted to Log\<LogData> when serializing

crates/primitives/src/log.rs
```diff
@@ -116,21 +116,49 @@ impl Log {
...
+        this.data.topics.encode(out);
+        this.data.data.encode(out);
     }
 
     fn length(&self) -> usize {
+        let this = self.reserialze();
         let payload_length =
-            self.address.length() + self.data.data.length() + self.data.topics.length();
+            this.address.length() + this.data.data.length() + this.data.topics.length();
         payload_length + alloy_rlp::length_of_length(payload_length)
```
Rjected, line 162: this might need to be `payload_length + alloy_rlp::length_of_length(payload_length) + 1`? or `Header { list: true, payload_length }.length() + payload_length`, would check length assert tho

Rjected (changes requested):
mostly nits, otherwise lgtm

crates/primitives/src/log.rs
prestwich: length assert in latest commit works fine

crates/primitives/src/log.rs
prestwich: right. and because we are _usually_ 

This is significantly less ergonomic with `DynSolEvent` tho, as a decoded sol event can't be assumed to be validly encoded the way a decoded `sol!` struct with `SolEvent` can be. That 
assumption allows the infallible reserialize

cc @DaniPopes do you think that we should attach the `DynSolEvent` as a prop on the `DecodedEvent` in dyn abi so that we can assume valid construction and encoding? This would let us `impl From<&DecodedEvent> for LogData`

prestwich:
cc @DaniPopes what's up with the tests?

DaniPopes:
You can update them by running with env `TRYBUILD=overwrite`

crates/primitives/src/log.rs
Rjected: ah, forgot that the method for header just inlined `length_of_length`

Rjected (commented):
This looks good to me

Rjected (approved):

Rjected:
this also needs to be rebased

prestwich:
clippy issue seems to have been introduced by merge of main. why?

Rjected:
> clippy issue seems to have been introduced by merge of main. why?
> 
> Please do not merge until we know why this is the case

the new nightly clippy doesnt like things that are never constructed, including in tests

onbjerg:
hmm unsure, was this struct ever used? doing

```
git grep "StructProp" (git rev-list --all)
```

yields nothing except changelog and the struct def itself

weird that it wouldn't be caught before this?

prestwich:
> the new nightly clippy doesnt like things that are never constructed, including in tests

why are we using nightly clippy? didnt we decide that was a mistake with ethers as it inteoduced unstable lints?

prestwich:
CI is unrelated.

DaniPopes:
This breaks RLP encoding for `Log<LogData>` (the default).

PR #563, opened 2024-03-11
feat(sol-macro): add event filters to contracts

DaniPopes:
Companion to https://github.com/alloy-rs/alloy/pull/274

Example usage: https://github.com/alloy-rs/alloy/blob/14a797b42cbfbdc857f0b0265e7c562df63d7f98/crates/contract/src/event.rs#L165

### DIFF
--- a/crates/sol-macro/src/expand/contract.rs
+++ b/crates/sol-macro/src/expand/contract.rs
@@ -450,7 +450,7 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
                 pub fn event_filter<E: alloy_sol_types::SolEvent>(&self, call: &C)
                     -> alloy_contract::Event<N, T, &P, E>
                 {
-                    alloy_contract::Event::new_sol(&self.provider, self.address)
+                    alloy_contract::Event::new_sol(&self.provider, &self.address)
                 }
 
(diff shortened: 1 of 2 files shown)


crates/sol-macro/src/expand/contract.rs
```diff
@@ -332,14 +332,29 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
...
+            let name = format_ident!("{name}_filter");
+            let doc = format!(
+                "Creates a new event filter for the [`{event_name}`] event.",
+            );
+            quote! {
+                #[doc = #doc]
+                pub fn #name(&self) -> alloy_contract::Event<N, T, &P, #event_name> {
+                    self.event_filter::<#event_name>()
+                }
+            }
```
gakonst, line 346: Should we make these default snake case? I realize that for `function doStuff` we codegen `foo.doStuff().send()` atm

gakonst (approved):
supportive once we have it in

crates/sol-macro/src/expand/contract.rs
DaniPopes: Currently we default everything to the same name as declared in solidity/ABI. It's possible in the future to implement renaming like in serde but currently there is no real demand.

PR #569, opened 2024-03-13
fix: force clippy to stable

### LINKED ISSUE #566: [Bug] Use stable clippy instead of nightly
### Component

Other (please provide more details)

_No response_

### Operating System

None

### Describe the bug

Nightly clippy includes unstable and changing lints and results in random CI breakage

prestwich:
## Motivation

Prevent future random CI breakage

closes #566 

## Solution

Set clippy to stable in CI

there may be a better way to achieve this, but this way is straightforward

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
-      - run: cargo clippy --workspace --all-targets --all-features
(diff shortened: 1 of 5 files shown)


mattsse (approved):
sounds reasonable to me.
no more false positive unstable issues

wdyt @DaniPopes

gakonst (approved):
I am OK with this

PR #577, opened 2024-03-17
fix(dyn-abi): correctly parse uints in `coerce_str`

ay:
## Motivation

When using `coerce_str` to parse a uint from a decimal number in non-wei units, precision is lost due to use of `f64`. For example, parsing `123456.123456789123456789 ether` returns `123456123456789126578176` wei, but the correct value is `123456123456789123456789` wei.

## Solution

I removed uses of `f64`, and all math is now done using `U256` only. This also means `std` is no longer required when parsing decimal numbers with `coerce_str`.

I also added test cases that fail without this change.

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
lgtm, thanks!

PR #578, opened 2024-03-17
refactor: break SolInput to its own crate

### LINKED ISSUE #567: [Refactor] Break `sol_macro::SolInput` into a new crate
### Component

sol! macro

Allow for re-use of the input in other proc macros

### Additional context

_No response_

prestwich:
## Motivation

Closes #567 

Decompose `sol!` macro and allow reuse of input across multiple expanders

## Solution

- Move `SolInput` and related
- Make an `Expander` trait

followup #579 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/sol-macro-input/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+mod input;
+pub use input::{SolInput, SolInputKind};
+
+mod expander;
+pub use expander::SolInputExpander;
+
+#[cfg(feature = "json")]
+mod json;
+
+extern crate syn_solidity as ast;
```
mattsse, line 30: move above mods?

mattsse (approved):
I don't have enough context for why this is beneficial, but it sounds reasonable to me

pending @DaniPopes

prestwich:
> I don't have enough context for why this is beneficial, but it sounds reasonable to me

incremental progress towards reusing the same input for multiple procmacros. Basically, this would allow us to ingest the solid input once, handle attributes, and re-names, and then run it through multiple expanders

i.e. We could separate the sol interface macro expansion logic, from the rpc bindings macro expansion logic, and then compose those expanders. Much cleaner, macro architecture, long-term.

onbjerg (approved):
supportive

PR #587, opened 2024-03-27
Rearranged type param order so that the Network param is the last

moricho:
## Motivation
To default Network parameters to Ethereum in `alloy-provider` and `alloy-contract`, they need to be moved to the last of the type params: https://github.com/alloy-rs/alloy/pull/356
According to those changes, we need to adjust the type param order in `sol!` as well.

from: https://github.com/alloy-rs/alloy/issues/263

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro/src/expand/contract.rs
+++ b/crates/sol-macro/src/expand/contract.rs
@@ -367,7 +367,7 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, contract: &ItemContract) -> Result<TokenS
 
             #[doc = #struct_doc]
             #[derive(Clone)]
-            pub struct #name<T, P, N> {
+            pub struct #name<T, P, N = alloy_contract::private::Ethereum> {
                 address: alloy_sol_types::private::Address,
                 provider: P,
                 _network_transport: ::core::marker::PhantomData<(N, T)>,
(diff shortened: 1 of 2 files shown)


moricho:
@onbjerg Thanks for your advice in https://github.com/alloy-rs/alloy/pull/356#issuecomment-2021855378. I've opened this PR to adjust the type param order in `sol!`

mattsse (approved):
that makes sense,

does this introduce more issues with patched alloy-core deps on alloy?

mattsse (commented):
this currently clashes with alloy-provider, no?

https://github.com/alloy-rs/alloy/blob/c22b9ddd6461793f7bf05d5bb32a3c0fb2b8fe71/crates/provider/src/provider.rs#L198-L198

so this is a alloy-provider issue first

DaniPopes (approved):
Thanks! I made the generated struct also default to `N = Ethereum`. Will merge once companion is merged first.

moricho:
@mattsse 

> does this introduce more issues with patched alloy-core deps on alloy?

This PR will only affect `alloy-contract` and `alloy-provider` and it's already covered by [the PR](https://github.com/alloy-rs/alloy/pull/356) I created in `alloy-rs/alloy`.
(`sol!` is being used only for tests in the packages for now)

moricho:
@DaniPopes Thanks! I've just fixed https://github.com/alloy-rs/alloy/pull/356 and it's ready now. Once this PR is merged, I'll update alloy dependencies to use the latest alloy-core

PR #592, opened 2024-04-04
Make topics pub in Log so we can implement Compact codec in reth_codec

0xqd:
Ref https://github.com/paradigmxyz/reth/pull/7474
@mattsse


DaniPopes:
The field is private intentionally, use the getter methods instead.

PR #596, opened 2024-04-09
feat(json-abi): support legacy JSON ABIs

DaniPopes:
Supersedes https://github.com/alloy-rs/core/pull/350
Closes https://github.com/foundry-rs/foundry/issues/7593


crates/json-abi/src/state_mutability.rs
```diff
@@ -0,0 +1,183 @@
...
+/// In particular, this will deserialize the `stateMutability` field if it is present,
+/// and otherwise fall back to the deprecated `constant` and `payable` fields.
+///
+/// Since it must be used in combination with `#[serde(flatten)]`, a `serialize` implementation
+/// is also provided, which will always serialize the `stateMutability` field.
+///
+/// # Examples
+///
+/// Usage: `#[serde(default, flatten, with = "serde_state_mutability_compat")]` on a
+/// [`StateMutability`] struct field.
```
mattsse, line 66: TIL this is possible

mattsse (approved):
lgtm

crates/json-abi/src/state_mutability.rs
DaniPopes: I was praying this would work, it did :pray:

PR #607, opened 2024-04-19
Add `SignatureError` variants

tcoratger:
Add `SignatureError` variants to address potential errors during conversion from a RPC transaction to a `TransactionSignedEcRecovered`.

Related: https://github.com/paradigmxyz/reth/pull/7752

### DIFF
--- a/crates/primitives/src/signature/error.rs
+++ b/crates/primitives/src/signature/error.rs
@@ -13,6 +13,12 @@ pub enum SignatureError {
     /// Invalid parity.
     InvalidParity(u64),
 
+    /// Missing parity.
+    MissingParity,
+
+    /// Invalid signature.
+    InvalidSignature,
+
     /// k256 error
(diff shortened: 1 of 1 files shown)


DaniPopes:
Sorry these don't belong here if the failure is in Reth code.

tcoratger:
> Add `SignatureError` variants to address potential errors during conversion from a RPC transaction to a `TransactionSignedEcRecovered`.
> 
> Related: [paradigmxyz/reth#7752](https://github.com/paradigmxyz/reth/pull/7752)

Sorry, as this error variant https://github.com/alloy-rs/alloy/blob/c7a51e9b20e8d136572c30fc889799d97f50bd02/crates/rpc-types/src/eth/transaction/error.rs#L14 wraps a `SignatureError`, I was thinking this is a good place to add some signature related error variants. But understood, just provided the equivalent error variants directly inside `ConversionError`.

Issue #588, opened 2024-03-28
[Feature] Ability to add documentation inside of sol! macro to satisfy `missing_docs` lint rule


zerosnacks:
### Component

sol! macro

### Describe the feature you would like

When using the following common lint rule:

```toml
[lints]
rust.missing_docs = "warn"
```

It will warn users that this `sol!` block is missing documentation with the error: `missing documentation for a struct field
requested on the command line with "-W missing-docs"`

```rust
sol! {
    #[sol(rpc, bytecode="6080...0033")]
    contract Counter {
        uint256 public number;

        function setNumber(uint256 newNumber) public {
            number = newNumber;
        }

        function increment() public {
            number++;
        }
    }
}
```

It is however currently not possible to add documentation inside of the `sol!` macro and therefore it requires adding a `#[allow(missing_docs)]` for now.

### Additional context

Added a minimal repro here: https://github.com/zerosnacks/alloy-core-repro-588

Unclear if this should be marked as a feature request or a bug

alexfertel:
Came here to open this exact issue!

Fwiw, this invocation gives the same warning:

```rust
sol! {
    /// Emitted when `value` tokens are moved from one account (`from`) to
    /// another (`to`).
    ///
    /// Note that `value` may be zero.
    event Transfer(address indexed from, address indexed to, uint256 value);
    /// Emitted when the allowance of a `spender` for an `owner` is set by a
    /// call to `approve`. `value` is the new allowance.
    event Approval(address indexed owner, address indexed spender, uint256 value);
}
```

DaniPopes:
Is it my understanding that you would prefer something like the following, instead of allowing the lint with `#[allow(missing_docs)]`?

```rust
sol! {
    /// ...
    event Approval(
        /// ...
        address indexed owner,
        /// ...
        address indexed spender,
        /// ...
        uint256 value
    );
}
```

alexfertel:
> Is it my understanding that you would prefer something like the following

For me this gives a warning as well.

DaniPopes:
Yes, it's not implemented, I am asking if that's what you would like

alexfertel:
Yeah, I think that works 👍

alexfertel:
> Is it my understanding that you would prefer something like the following, instead of allowing the lint with #[allow(missing_docs)]

Ah @DaniPopes, so I was revisiting this and reading the code, and I realized that I probably misunderstood your question twice.

[...]

PR #619, opened 2024-05-05
feat(sol-macro): allow missing docs for event fields

alexfertel:
See https://github.com/alloy-rs/core/issues/588#issuecomment-2094823588

## Motivation

We want this `sol!` invocation to not generate a `missing_docs` warning:

```rust
sol! {
    /// Emitted when `value` tokens are moved from one account (`from`) to
    /// another (`to`).
    ///
    /// Note that `value` may be zero.
    event Transfer(address indexed from, address indexed to, uint256 value);
    /// Emitted when the allowance of a `spender` for an `owner` is set by a
    /// call to `approve`. `value` is the new allowance.
    event Approval(address indexed owner, address indexed spender, uint256 value);
}
```

## Solution

We add an `#[allow(missing_docs)]` annotation to the generated event struct fields. The alternative would be supporting:

```rust
sol! {
    /// ...
    event Approval(
        /// ...
        address indexed owner,
        /// ...
        address indexed spender,
        /// ...
        uint256 value
    );
}
```

Which is less readable and redundant.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

I don't think the above apply, but lmk if that is not the case.

### DIFF
--- a/crates/sol-macro/src/expand/event.rs
+++ b/crates/sol-macro/src/expand/event.rs
@@ -139,7 +139,10 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, event: &ItemEvent) -> Result<TokenStream>
         #[allow(non_camel_case_types, non_snake_case, clippy::style)]
         #[derive(Clone)]
         pub struct #name {
-            #(pub #fields,)*
+            #(
+                #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
lgtm

PR #632, opened 2024-05-22
feature: DynSolCall

prestwich:
## Motivation

Continue fleshing out dyn-abi to reach feature parity with sol-types

## Solution

- Add `DynSolCall`
- Add `DynSolReturns`
- Add `impl Specifier<DynSolCall> for Function`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/dyn-abi/src/eip712/mod.rs
+++ b/crates/dyn-abi/src/eip712/mod.rs
@@ -16,3 +16,39 @@ mod resolver;
 pub use resolver::{PropertyDef, Resolver, TypeDef};
 
 pub(crate) mod coerce;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use alloy_primitives::B256;
+    use alloy_sol_types::SolStruct;
+    #[test]
+    fn repro_i128() {
+        alloy_sol_types::sol! {
+            #[derive(serde::Serialize)]
+            struct Order {
+                bytes32 sender;
+                int128 priceX18;
+                int128 amount;
+                uint64 expiration;
+                uint64 nonce;
+            }
+        }
+
+        let msg = Order {
+            sender: B256::repeat_byte(3),
+            priceX18: -1 * 1000000000000000,
+            amount: 2,
+            expiration: 3,
+            nonce: 4,
+        };
+
+        let domain = Default::default();
+
+        let static_sh = msg.eip712_signing_hash(&domain);
+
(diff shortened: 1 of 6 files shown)


prestwich:
anyone know what's up with miri cancelling?

crates/dyn-abi/src/eip712/mod.rs
```diff
@@ -16,3 +16,39 @@ mod resolver;
 pub use resolver::{PropertyDef, Resolver, TypeDef};
 
 pub(crate) mod coerce;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use alloy_primitives::B256;
+    use alloy_sol_types::SolStruct;
+    #[test]
```
DaniPopes, line 25:
```suggestion

    #[test]
```

crates/dyn-abi/src/ext/abi.rs
```diff
@@ -118,29 +118,29 @@ impl JsonAbiExt for Error {
...
     #[inline]
     fn abi_encode_output(&self, values: &[DynSolValue]) -> Result<Vec<u8>> {
-        encode_typeck(&self.outputs, values)
+        self.resolve()?.abi_encode_output(values)
     }
 
     #[inline]
     fn abi_decode_output(&self, data: &[u8], validate: bool) -> Result<Vec<DynSolValue>> {
-        abi_decode(data, &self.outputs, validate)
+        self.resolve()?.abi_decode_output(data, validate)
```
DaniPopes, line 143: Can you please keep these specialized functions, we don't want to calculate the signature and collect inputs and outputs on every call

DaniPopes (changes requested):

crates/dyn-abi/src/ext/abi.rs
prestwich: hmmm. I think that if users are keeping the json representation in memory and passing it around, that's generally an anti-pattern, right?  I see these traits as convenience functions for one-off things and to patch functionality holes, not as core API. `Dyn____` is our in-memory representation of ABI information, and greedily resolving a `JsonAbi` into a `DynAbi` is the longterm officially-recommended path

prestwich: reverting the file to unblock this

DaniPopes (approved):

Issue #630, opened 2024-05-17
[Bug] Panic in `Parity::inverted`


prestwich:
### Component

primitives

### Operating System

None

### Describe the bug

This line needs to handle the 0 case
https://github.com/alloy-rs/core/blob/main/crates/primitives/src/signature/parity.rs#L109

jafar75:
I can send a PR for this issue
just to be ensured, for 0 case, it must return 0?

prestwich:
for `inverted(0)` should return `1`. even tho this means the `inverted` operation is no longer symmetrical. we don't expect real users to ever hit this edge case

PR #633, opened 2024-05-23
fix: handle 0 for inverting eip155 parity.

jafar75:
## Motivation

As mentioned in #630 , there is a corner case for inverting parity with value of Eip155(0). Program will try to return Eip155(-1) which is panic issue. 

## Solution

One should handle the 0 case separately and return Eip155(1) as the inverted output.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


prestwich:
CI is unrelated

Issue #240, opened 2023-08-23
Prevent user from instantiating invalid `Param` instances by making at least 1 field private


prestwich:
if public fields allow the user to create invalid instances, then we should at least consider making them non-public

_Originally posted by @prestwich in https://github.com/alloy-rs/core/pull/233#discussion_r1300362251_

chirag-bgh:
@prestwich I can take this.

prestwich:
ah my guide above is incorrect. because `name` and `type` need to be validated, they need to be private. no `_guard` needed

@DaniPopes has pointed out that this breaks mathcing and destructuring, so we should evaluate how worth it it is after seeing code. so there's a chance even if you do this work it won't get merged

mahmudsudo:
can i take this ?

PR #634, opened 2024-05-23
feat: Create new method on Param and EventParam

quentinv72:
## Motivation

Resolve https://github.com/alloy-rs/core/issues/240

## Solution

* Create a `new` method for `Param` and `EventParam` 
* Annotate the `ty` and `name` fields on `Param` and `EventParam` with `#[doc(hidden)]` to discourage users from creating their own isntances of those types wihtout using the `new` method. The reason for not making those fields private is that they are used by macros in different crates.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


prestwich:
Thank you, I like this approach

PR #639, opened 2024-05-30
Fix Log serde for non self describing protocols

kpp:
Log flattens data for serde, which is an issue for a non self describing protocols as bcs and bincode. Without this fix test will fail with:

1. bcs:

```
---- log::tests::test_bcs_encode_decode stdout ----
thread 'log::tests::test_bcs_encode_decode' panicked at crates/primitives/src/log.rs:248:51:
called `Result::unwrap()` on an `Err` value: NotSupported("deserialize_any")
```

2. bincode:

```
---- log::tests::test_bincode_encode_decode stdout ----
thread 'log::tests::test_bincode_encode_decode' panicked at crates/primitives/src/log.rs:227:52:
called `Result::unwrap()` on an `Err` value: SequenceMustHaveLength
```

Code affected: reth_primitives::Receipt.

### DIFF
--- a/crates/primitives/src/log.rs
+++ b/crates/primitives/src/log.rs
@@ -83,7 +83,6 @@ impl LogData {
 
 /// A log consists of an address, and some log data.
 #[derive(Clone, Default, Debug, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(derive_arbitrary::Arbitrary, proptest_derive::Arbitrary))]
 pub struct Log<T = LogData> {
     /// The address which emitted this log.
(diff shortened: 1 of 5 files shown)


crates/primitives/src/log.rs
```diff
@@ -89,7 +89,6 @@ pub struct Log<T = LogData> {
     /// The address which emitted this log.
     pub address: Address,
     /// The log data.
-    #[cfg_attr(feature = "serde", serde(flatten))]
     pub data: T,
```
prestwich, line 92: this will now produce incorrect results for ethereum json rpc . supporting standard ethereum json rpc is a primary goal of alloy

prestwich: that would prevent us from being generic over decoded log structs

kpp: What is your suggested solution?

prestwich: for us, the current behavior is working as intended. we don't have an intended solution as serializing logs to any format other than JSON is not a goal of the library

kpp: Will you accept a solution that preserves the current JSON serialization but fixes binary serde?

prestwich: yes, we would

kpp: Done

crates/primitives/src/log_serde.rs
```diff
@@ -0,0 +1,128 @@
+//! This is an implementation of serde for Log for
+//! both human-readable and binary forms.
+//!
+//! Etherium JSON RPC requires logs in a flattened form.
```
DaniPopes, line 4:
```suggestion
//! Ethereum JSON RPC requires logs in a flattened form.
```

prestwich (approved):

PR #648, opened 2024-06-04
feat(sol-macro): add return value names to simple getters

DaniPopes:
Example:
```rust
sol! {
    contract Counter {
        uint256 public number;
    }
}

// Before:
Counter::numberReturn { _0: U256 }
// After:
Counter::numberReturn { number: U256 }
```

See the added test


mattsse (approved):

DaniPopes:
This mismatches Solc-emitted ABI because it does not populate return names

Issue #653, opened 2024-06-10
[Feature] Add additional aliases


zerosnacks:
### Component

Other (please provide more details)

### Describe the feature you would like

In an effort to reduce magic values in `alloy` I suggest adding two more aliases:

- `TxNonce`
- `BlockTimestamp`

Related PR: https://github.com/alloy-rs/alloy/pull/859

PR #654, opened 2024-06-10
feat(primitives): add additional common aliases

zerosnacks:
## Motivation

Closes: https://github.com/alloy-rs/core/issues/653

For context + related PR in `alloy`: https://github.com/alloy-rs/alloy/pull/859

## Solution

Add generic aliases unrelated to specific EIPs

- `TxNonce` (with doc alias)
- `BlockTimestamp`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/aliases.rs
+++ b/crates/primitives/src/aliases.rs
@@ -65,6 +65,9 @@ pub type BlockHash = B256;
 /// A block number.
 pub type BlockNumber = u64;
 
+/// A block timestamp.
+pub type BlockTimestamp = u64;
+
 /// A transaction hash is a keccak hash of an RLP encoded signed transaction.
 #[doc(alias = "TransactionHash")]
 pub type TxHash = B256;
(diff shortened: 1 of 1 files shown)


crates/primitives/src/aliases.rs
```diff
@@ -65,6 +65,9 @@ pub type BlockHash = B256;
 /// A block number.
 pub type BlockNumber = u64;
 
+/// A block timestamp.
+pub type BlockTimestamp = u64;
```
onbjerg, line 69: unsure if we should add this w/o also using it, so maybe a separate pr?

onbjerg (approved):
lgtm otherwise, no strong feeling either way, just dont want us to end up not using the new alias:)

crates/primitives/src/aliases.rs
zerosnacks: I would update https://github.com/alloy-rs/alloy/pull/859 to include it

Same for `TxNonce`

zerosnacks:
> hmm, I'm sceptical that these aliases are useful

Added some additional context here: https://github.com/alloy-rs/alloy/pull/859

>Reduces magic values, reduces chance of error and makes it easier for end-users to see what value they are getting e.g. TxHash instead of FixedBytes<32> when watch()-ing a transaction. Main goal is to limit showing internal types and help with consistency / standardize fields (`u64` vs `U64`) and sizes.

DaniPopes (approved):

PR #661, opened 2024-06-13
feat(primitives): serialize yParity and v simultaneously

yash-atreya:
## Motivation

Previously, when `v` was set to `0x0 || 0x1` i.e `Parity::Parity(false || true)`, only `yParity` would appear in the serialized form of the sig.

## Solution

Serialize `v` along with `yParity` in that case. 
Ref: https://github.com/ethereum/go-ethereum/blob/473ee8fc07a3f89cf3978e164a6fad218de9a6b5/internal/ethapi/api.go#L1342-L1345 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -418,8 +418,14 @@ impl serde::Serialize for crate::Signature {
             match self.v {
                 Parity::Eip155(v) => map.serialize_entry("v", &crate::U64::from(v))?,
                 Parity::NonEip155(b) => map.serialize_entry("v", &(b as u8 + 27))?,
-                Parity::Parity(true) => map.serialize_entry("yParity", "0x1")?,
-                Parity::Parity(false) => map.serialize_entry("yParity", "0x0")?,
+                Parity::Parity(true) => {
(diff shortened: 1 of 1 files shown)


crates/primitives/src/signature/sig.rs
```diff
@@ -418,8 +418,14 @@ impl serde::Serialize for crate::Signature {
...
                 Parity::NonEip155(b) => map.serialize_entry("v", &(b as u8 + 27))?,
-                Parity::Parity(true) => map.serialize_entry("yParity", "0x1")?,
-                Parity::Parity(false) => map.serialize_entry("yParity", "0x0")?,
+                Parity::Parity(true) => {
+                    map.serialize_entry("v", "0x1")?;
+                    map.serialize_entry("yParity", "0x1")?;
+                }
+                Parity::Parity(false) => {
+                    map.serialize_entry("v", "0x0")?;
+                    map.serialize_entry("yParity", "0x0")?;
```
DaniPopes, line 427: can be merged in one branch with `Parity(parity) => { let value = if parity {} else {} }`

DaniPopes (changes requested):

PR #664, opened 2024-06-14
feat(dyn-abi): derive Default for TypedData

DaniPopes:
cc @prestwich is this right or does this need a custom impl to handle resolver/..?

### DIFF
--- a/crates/dyn-abi/src/eip712/typed_data.rs
+++ b/crates/dyn-abi/src/eip712/typed_data.rs
@@ -63,7 +63,7 @@ impl<'de> Deserialize<'de> for Eip712Types {
 ///     "required": ["types", "primaryType", "domain", "message"]
 /// }
 /// ```
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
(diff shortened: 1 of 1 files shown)


prestwich (changes requested):
i think its never correct to have an empty primary type or an empty value. those would both be invalid instances

DaniPopes:
What's a sensible default for no-op / testing then?

mattsse (commented):
I think it'd be okay if default is always invalid

prestwich:
> I think it'd be okay if default is always invalid

I would be very against this

why not just add a `cfg_attr(test, derive(Default))`

mattsse:
this only works internally I think

prestwich:
instead of proposing a feature, can you talk about the need? what are you trying to test?

DaniPopes:
dummy input to eth_signTypedData

prestwich:
sure but do you want input that's going to cause an error, GIGO, or something else?

gakonst:
if useful for testing we have used `test-utils` features in the past instead of `cfg_attr(test)`, and i'd be ok w it

PR #668, opened 2024-06-16
[BugFix] Use Cow<'a, str> to be compatible with serde_json::from_reader

### LINKED ISSUE #667: [Bug] Incompatible with `serde_json::from_reader`
### Component

json-abi, primitives

0.7.5

### Operating System

Linux

### Describe the bug

See reproduction below:

```rust
use std::io::Write;
use alloy_json_abi::Function;

fn main() {
        let s = r#"{"type":"function","name":"increaseAllowance","inputs":[{"name":"spender","type":"address","internalType":"address"},{"name":"addedAmount","type":"uint256","internalType":"uint256"}],"outputs":[{"name":"","type":"bool","internalType":"bool"}],"stateMutability":"nonpayable"}"#;
        let f: Function = serde_json::from_str(s).unwrap(); // works

        let mut fp = std::fs::OpenOptions::new().write(true).create(true).truncate(true).open("/tmp/f.json").unwrap();
        fp.write_all(s.as_bytes()).unwrap();

        let mut fp = std::fs::OpenOptions::new().read(true).open("/tmp/f.json").unwrap();
        
        let f: Function = serde_json::from_reader(fp).unwrap(); // not working
}
```

The root cause is `alloy` trying to firstly deserialize the content into a borrowed type [here](https://github.com/alloy-rs/core/blob/b76881b02f1e11622e1266142740c6dc20f14e6b/crates/json-abi/src/param.rs#L61), which is generally not compatible with `serde_json::from_reader`. 

The possible solution is to use `Cow<'a, str>` to replace `&'a str` as mentioned in https://github.com/serde-rs/serde/issues/914 , https://github.com/rust-analyzer/smol_str/issues/14 and https://github.com/serde-rs/serde/issues/914 . However, I ran into lifetime issues, so I leave an issue here first.

wtdcode:
## Motivation

Fix #667

## Solution

As title, use `Cow<'a, str>` to be compatible with `serde_json::from_reader`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

I don't see any possible breaking changes because previously it is not working. Besides, it shouldn't incur any performance penalty because from general `from_str` the data is still borrowed.

### DIFF
--- a/crates/json-abi/src/param.rs
+++ b/crates/json-abi/src/param.rs
@@ -8,7 +8,7 @@ use alloc::{
     string::String,
     vec::Vec,
 };
-use core::{fmt, str::FromStr};
+use core::{borrow::Borrow, fmt, str::FromStr};
 use parser::{Error, ParameterSpecifier, TypeSpecifier};
 use serde::{de::Unexpected, Deserialize, Deserializer, Serialize, Serializer};
 
@@ -62,8 +62,8 @@ impl<'de> Deserialize<'de> for Param {
             if inner.indexed.is_none() {
                 inner.validate_fields()?;
                 Ok(Self {
-                    name: inner.name.to_owned(),
-                    ty: inner.ty.to_owned(),
+                    name: inner.name.into_owned(),
+                    ty: inner.ty.into_owned(),
                     internal_type: inner.internal_type.map(Into::into),
                     components: inner.components.into_owned(),
                 })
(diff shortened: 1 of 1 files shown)


wtdcode:
Found this:

```
    fn visit_str<E>(self, _v: &str) -> Result<Self::Value, E>
    where
        E: serde::de::Error,
    {
        // `from_reader` copies the bytes into a Vec before calling this
        // method. Because the lifetime is unspecified, we can't borrow from it.
        // As a result, we don't support `from_reader`.
        Err(serde::de::Error::custom(
            "Using serde_json::from_reader is not supported. Instead, buffer the reader contents into a string, as in alloy_json_abi::JsonAbi::load.",
        ))
    }
```

Looks like I need to fix more code.

wtdcode:
Give up.

PR #689, opened 2024-07-19
feat: expose Signature<T>

leruaa:
## Motivation

Needed for https://github.com/alloy-rs/alloy/issues/1077 and https://github.com/paradigmxyz/reth/pull/9593

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -60,5 +60,3 @@ arbitrary = [
 ]
 k256 = ["alloy-primitives/k256"]
 eip712 = ["alloy-sol-types?/eip712-serde", "alloy-dyn-abi?/eip712"]
-
-unstable-doc = ["alloy-primitives/unstable-doc"]
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -12,11 +12,6 @@ pub use alloy_primitives as primitives;
 #[doc(no_inline)]
 pub use primitives::{hex, uint};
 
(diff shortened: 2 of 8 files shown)


leruaa:
While attempting to apply this refactoring to alloy itself, I start to understand why the `k256` feature is not additive. That to avoid to impl traits like SignableTransaction for multiple Signature types leading to the "multiple `impl`s ... found" error.

I'm trying to find a workaround.

mattsse:
@leruaa perhaps we can introduce a helper trait in alloy for the required rsv function and just that for the alloy encode signature fns instead?

mattsse (commented):
this would introduce a ton of friction

I agree the memoized type is a bit horribly, but we can probably solve our problems alloy internally by introducing a signature encodeable trait there so we can implement it for reth's signature type as well. 
can deal with the generic signature snafu later

mattsse:
but perhaps we should consider removing the generic entirely, I find this very useless because this is effectively just duplicating the rsv again

https://github.com/alloy-rs/core/blob/adcf7adfa1f35c56e6331bab85b8c56d32a465f1/crates/primitives/src/signature/sig.rs#L301-L305

leruaa:
@mattsse fyi and for posterity, the branch in alloy applying this refactoring in alloy-core: https://github.com/alloy-rs/alloy/compare/main...leruaa:alloy:signature_refactoring

Issue #660, opened 2024-06-13
[Bug] Type paths lost when generating code with sol! macro via json abi


benluelo:
### Component

json-abi, sol! macro

### What version of Alloy are you on?

v0.7.6

### Operating System

Linux

### Describe the bug

If a struct is in a library, the library path is lost when generating code from the json abi.

Take the following solidity code:

```solidity
library LibA {
    struct Struct {
        uint64 field64;
    }
}

library LibB {
    struct Struct {
        uint128 field128;
    }
}

contract Contract {
    LibA.Struct internal aValue;
    LibB.Struct internal bValue;

    constructor(
        LibA.Struct memory aValue_,
        LibB.Struct memory bValue_
    )
    {
        aValue = aValue_;
        bValue = bValue_;
    }

    function fn(
        LibA.Struct memory aValue_,
        LibB.Struct memory bValue_
    ) public
    {
        aValue = aValue_;
        bValue = bValue_;
    }
}
```

this generates the following (abridged) rust code:

````rust
/**
Generated by the following Solidity interface...
```solidity
interface Contract {
    struct Struct {
        uint64 field64;
    }

    constructor(Struct aValue_, Struct bValue_);

    function fn(Struct memory aValue_, Struct memory bValue_) external;
}
```
*/
pub mod Contract {
    use ::alloy_sol_types as alloy_sol_types;
    /**```solidity
    struct Struct { uint64 field64; }
    ```*/
    pub struct Struct {
        pub field64: u64,
    }
    /**Constructor`.
    ```solidity
    constructor(Struct aValue_, Struct bValue_);
    ```*/
    pub struct constructorCall {
        pub aValue_: <Struct as ::alloy_sol_types::SolType>::RustType,
        pub bValue_: <Struct as ::alloy_sol_types::SolType>::RustType,
    }
    /**Function with signature `fn((uint64),(uint64))` and selector `0xd7d61df0`.
    ```solidity
    function r#fn(Struct memory aValue_, Struct memory bValue_) external;
    ```*/
    pub struct fnCall {
        pub aValue_: <Struct as ::alloy_sol_types::SolType>::RustType,
        pub bValue_: <Struct as ::alloy_sol_types::SolType>::RustType,
    }
    pub struct fnReturn {}
    ///Container for all the [`Contract`](self) function calls.
    pub enum ContractCalls {
        r#fn(fnCall),
    }
}
````

[...]

PR #690, opened 2024-07-23
fix: correctly resolve duplicate struct names in json abi

klkvr:
## Motivation

Closes https://github.com/alloy-rs/core/issues/660

## Solution

I've added second abi preprocessing method `JsonAbi::rename_internal_structs`, which goes over ABI, resolves structs that have a same name but different components sets, and renames them into `<struct name>0`, `<struct name>1`, etc

It should be also possible to key them by `contract` instead, however, there are some edge cases related to aliases with such approach. E.g. `import {Struct as Struct1} from "..."; import {Struct as Struct2} from "...";` will result in two different structs referenced as `Struct` in ABI. Same is possible with `Lib.Struct` if different `Lib`s are aliased.

This edge case is kind of weird, so I am open to changing impl if we consider performance improvements worth it.

Also not sure if we want this to happen on the level of `JsonAbi::to_sol` instead of only for `sol!` used code

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/json-abi/src/abi.rs
```diff
@@ -608,3 +649,82 @@ impl<'de> Visitor<'de> for ContractObjectVisitor {
         Ok(ContractObject { abi, bytecode, deployed_bytecode })
     }
 }
+
+/// A [crate::Visitor] implementation that collects a mapping from struct name to set of components
+/// various structs with such name have.
+#[derive(Debug, Default, Clone)]
+struct Collector<'a> {
+    unique_struct_components: BTreeMap<&'a str, BTreeSet<Vec<&'a str>>>,
```
mattsse, line 657: perhaps worth using smallvec here for the `Vec` because we only expect a few projections?

crates/json-abi/src/abi.rs
```diff
@@ -246,6 +253,40 @@ impl JsonAbi {
...
+            .map(|(name, components)| {
+                (
+                    name.to_string(),
+                    components
+                        .into_iter()
+                        .enumerate()
+                        .map(move |(idx, components)| {
+                            let components =
+                                components.into_iter().map(ToString::to_string).collect::<Vec<_>>();
+                            (components, format!("{name}{idx}"))
```
mattsse, line 279: should we start at 0 here or at 1?

crates/json-abi/src/abi.rs
```diff
@@ -608,3 +649,82 @@ impl<'de> Visitor<'de> for ContractObjectVisitor {
         Ok(ContractObject { abi, bytecode, deployed_bytecode })
     }
 }
+
+/// A [crate::Visitor] implementation that collects a mapping from struct name to set of components
+/// various structs with such name have.
```
mattsse, line 654: ah okay, I understand how this works, first we traverse the abi and collect unique names and their paths

mattsse (changes requested):
cool, I believe this works and is easy to follow.

can we add another test case for https://github.com/foundry-rs/foundry/issues/8444

imo we can ignore this edge case `import {Struct as Struct2}`

ptal @DaniPopes

crates/json-abi/src/abi.rs
```diff
@@ -246,6 +253,40 @@ impl JsonAbi {
         }
     }
 
+    /// Goes over all structs referenced as internal types in the ABI and changes them to global
+    /// structs.
+    ///
+    /// Namespace for global struct names is managed and in cases of duplicates structs are renamed.
+    pub fn rename_internal_structs(&mut self) {
```
DaniPopes, line 260: I don't think any of this should be exposed publicly, it should only be a sol! implentation detail

DaniPopes (commented):
I don't really like renaming because this doesn't tackle the underlying problem that namespaces are not handled in sol!, e.g. this only fixes json input

I have a branch open where I make namespaced output work, but got stuck on weird compile errors https://github.com/alloy-rs/core/tree/dani/to-sol-contract-specifiers

crates/json-abi/src/abi.rs
klkvr: function overloads start with 0, so I've just made it similar

just noticed that functions are named as `function_0`, `function_1`, should we include underscore here as well?

klkvr:
> I have a branch open where I make namespaced output work, but got stuck on weird compile errors https://github.com/alloy-rs/core/tree/dani/to-sol-contract-specifiers

yeah, such approach indeed is more flexible

current branch compiles fine for me, do you have more fixes for sol! as wip?

from diff and a quick test it seems that updated `Type::Custom` branch in `rec_expand_type` does most of the job on the `sol!` side? would be happy to finish this if you'll point out what's left/not working yet

DaniPopes:
It fails to compile in tests in crates/sol-types/tests/macros/sol/json.rs, but expanding the output and compiling it separately is fine

It only handles the ToSol side, expanding libraries, but namespaces are not handled in the sol! internal representation, so function signatures are wrong

Probably best to just continue that branch

klkvr:
> It fails to compile in tests in crates/sol-types/tests/macros/sol/json.rs, but expanding the output and compiling it separately is fine

fyi moving `sol!` invocation out of the test function scope fixes compilation, will look into why it's happening

klkvr:
one of the issues seems to be a limitation of function-level modules

e.g. this code doesn't work
```rust
fn test_fn() {
    pub mod inner1 {
        pub struct InnerStruct {}
    }
    
    pub mod inner2 {
        use super::*;
        fn some_fn() {
            inner1::InnerStruct {};
        }
    }
}
```

and it seems that there's no way to reach other modules defined inside of a function from inside of another module

DaniPopes:
Superseded by https://github.com/alloy-rs/core/pull/693, #694

PR #692, opened 2024-07-24
feat: add EncodableSignature trait

leruaa:
## Motivation

Per @mattsse comment: https://github.com/alloy-rs/core/pull/689#issuecomment-2242290238

To give more context: the goal is to standardize signature encoding via an helper trait to be able to use alloy tx types in reth with reth signature type.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -70,7 +70,7 @@ criterion.workspace = true
 serde_json.workspace = true
 
 [features]
-default = ["std"]
+default = ["std", "k256"]
 std = [
     "bytes/std",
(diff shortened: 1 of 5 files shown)


mattsse (commented):
this seems like an appropriate solution to our problem, but I wonder could we introduce this trait in the alloy crate instead first?

everything that's required by this trait should be available via public API and could be implemented on the exported `Signature` type I believe?

leruaa:
Done in https://github.com/alloy-rs/alloy/pull/1100

PR #693, opened 2024-07-25
feat: implement namespaces on `sol!` side

klkvr:
## Motivation

Makes `sol!` macro respect namespaces.

## Solution

- `current_namespace` context variable is added to `ExpCtxt` which is switched when we enter/exit contract body during AST traversal or expansion.
- `all_items`, `custom_types` and `overloaded_items` are wrapped into `NamespacedMap` which resolves full `SolPath`s depending on current namespace

cc @DaniPopes please take a look and lmk if I should switch base to master instead

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
this should work,
pending @DaniPopes

DaniPopes (approved):
lgtm

Issue #599, opened 2024-04-13
[Bug] Cannot define functions with the same signature in different contracts/interfaces


makcandrov:
### Component

sol! macro

### What version of Alloy are you on?

0.7.0

### Operating System

None

### Describe the bug

```rust
alloy_sol_types::sol! {
    interface IERC20 {
        function transfer(address to, uint256 value) external returns (bool);
    }

    interface IERC721 {
        function transfer(address to, uint256 tokenId) external returns (bool);
    }
}
```

```
error: function with same name and parameter types defined twice
       
         = note: other declaration is here
       
 --> src\main.rs:3:18
  |
3 |         function transfer(address to, uint256 value) external returns (bool);
  |                  ^^^^^^^^

```

DaniPopes:
You should be able to work around this for now by creating separate `sol!` invocations for each interface.

makcandrov:
It doesn't work when you need to share something (like a structure) between the interfaces

```rust
alloy_sol_types::sol! {
    struct S {
        uint256 s;
    }

    interface Foo {
        function foo(S memory s) external;
    }
    
}

alloy_sol_types::sol! {
    interface Bar {
        function foo(S memory s) external;
    }
}
```

```
error: unresolved type
       
         = help: Custom types must be declared inside of the same scope they are referenced in,
       or "imported" as a UDT with `type ... is (...);`
       
  --> src\main.rs:14:22
   |
14 |         function foo(S memory s) external;
   |                      ^

error: unresolved custom type: S
  --> src\main.rs:14:22
   |
14 |         function foo(S memory s) external;
   |                      ^
```

PR #694, opened 2024-07-25
feat(sol-macro): support namespaces

DaniPopes:
Closes https://github.com/alloy-rs/core/issues/660
Closes https://github.com/alloy-rs/core/issues/599


klkvr:
just tested locally and it seems that derives aren't applied to generated libraries in json abi input causing compilation issues

e.g. this would cause `Lib::Struct doesn't implement Debug`
```rust
sol!(
    #[derive(Debug)]
    Contract,
    r#"[
        {
            "type": "function",
            "name": "f",
            "inputs": [
                {
                    "name": "s",
                    "type": "tuple",
                    "internalType": "struct Lib.Struct",
                    "components": [
                        {
                            "name": "a",
                            "type": "uint256",
                            "internalType": "uint256"
                        }
                    ]
                }
            ],
            "outputs": [],
            "stateMutability": "nonpayable"
        }
    ]"#
);
```

DaniPopes:
And would also be a pretty big breaking change

DaniPopes:
This is broken now, marking as draft again:
```sol
sol! {
    contract Vm {
        struct Wallet {
            uint stuff;
        }

        /// Gets the nonce of an account.
        function getNonce(address account) external view returns (uint64 nonce);

        /// Get the nonce of a `Wallet`.
        function getNonce(Wallet calldata wallet) external returns (uint64 nonce);
    }
}
```

klkvr:
Updated in 18c2ed4, we didn't apply namespaces when resolving overloads

PR #698, opened 2024-07-28
chore(primitives): re-use ruint mask function

### LINKED ISSUE #695: Why redefine this mask function when ruint crate already have same mask function?
https://github.com/alloy-rs/core/blob/adcf7adfa1f35c56e6331bab85b8c56d32a465f1/crates/primitives/src/signed/int.rs#L107

https://github.com/recmo/uint/blob/c4616bdb1bcdaa991ef23f9808c56c86127625ea/src/lib.rs#L321

rongyi:
the underline ruint crate already have mask method

## Motivation
DRY

## Solution
use ruint::mask

Fix #695

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
Thanks!

PR #701, opened 2024-07-30
Remove `ethereum_ssz` dependency

eserilev:
## Motivation

Lighthouse is currently migrating our code base and relevant dependencies to `alloy_primtives`. During the migration we've come across some circular dependency issues between the `ethereum_ssz` and `alloy_primitives` crates. 

## Solution

This PR aims to remove the optional `ethereum_ssz` dependency as well as the `ssz` feature flag.

I have an open PR [here](https://github.com/sigp/ethereum_ssz/pull/26) that implements ssz `Decode` and `Encode` for the alloy generic types `Bytes` and `FixedBytes<N>` in `ethereum_ssz`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
nice,

in order to transition we need to coordinate releases,
we first need to merge this pr, then do a new alloy-core release, followed by an ssz release so we can also update all dependent crates.

so we can merge this once ethereum_ssz is already ready to publish

DaniPopes (approved):

michaelsproul:
How about this for a release strategy?

1. Release `ethereum_ssz` v0.6 which depends on `alloy-primitives` 0.7.7
2. Release `alloy-primitives` v0.8 which does not depend on `ethereum_ssz` at all
3. Release `ethereum_ssz` v0.7 which depends on `alloy-primitives` v0.8

We can try to make the timing quite tight between (2) and (3). I will try to get (1) out ASAP.

DaniPopes:
Sounds good, although I'm not sure if you need (1), but up to you

PR #704, opened 2024-08-04
fix: correctly expand derives for namespaced JSON outputs

klkvr:
## Motivation

Addresses https://github.com/alloy-rs/core/pull/694#issuecomment-2249896421

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/sol-macro-input/src/json.rs
```diff
@@ -19,10 +19,39 @@ impl SolInput {
...
+
+        let library_tokens = library_tokens_iter.by_ref();
+
+        let mut libraries = Vec::new();
+
+        while library_tokens.peek().is_some() {
+            let sol_library_tokens: TokenStream = std::iter::once(TokenTree::Ident(id("library")))
+                .chain(
+                    library_tokens
+                        .take_while(|tt| !matches!(tt, TokenTree::Ident(id) if id == "library")),
```
mattsse, line 44: could this end up looping if `library_tokens.take_while` does not consume a value?

klkvr: It should just consume all left elements if closure never returns `true`

mattsse: right

mattsse (approved):
lgtm,

clippy is unrelated

DaniPopes (approved):

PR #707, opened 2024-08-07
feat: derive `Hash` for `Sealed`

klkvr:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -4,7 +4,7 @@ use crate::B256;
 ///
 /// We do not implement any specific hashing algorithm here. Instead types
 /// implement the [`Sealable`] trait to provide define their own hash.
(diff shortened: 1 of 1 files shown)


crates/primitives/src/sealed.rs
```diff
@@ -4,7 +4,7 @@ use crate::B256;
 ///
 /// We do not implement any specific hashing algorithm here. Instead types
 /// implement the [`Sealable`] trait to provide define their own hash.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
```
mattsse, line 7: we could impl this manually and use the hash, but since we allow unchecked instantiation, this makes sense

mattsse (approved):

PR #709, opened 2024-08-13
chore: derive default on `Sealed`

rkrasiuk:
## Description

Derive `Default` on `Sealed` to be able to represent `Sealed*` reth types.


mattsse (changes requested):
this is a massive footgun, because now you can create types that are sealed with the zero hash.

you can implement this manually where T: sealable and instantiate it like this

PR #710, opened 2024-08-13
feat: derive ser deser on `Sealed`

rkrasiuk:
## Description

Derive serde's `Serialize` and `Deserialize` on `Sealed`.


DaniPopes (approved):

mattsse (approved):
this should be okay, assume the hash we're deserializing is valid sounds reasonable to me

PR #719, opened 2024-09-02
refactor: remove `Signature` generic

klkvr:
## Motivation

Currently it is impossible to construct an invalid signature, because `k256` feature requires it to contain a valid `k256::Signature`. This makes it impossible to decode and store invalid signatures. 

For example, we require this in reth to keep deposit transaction signatures: https://github.com/paradigmxyz/reth/blob/42dc5eea1685385b835edd03029640609f4c4ab4/crates/primitives/src/transaction/signature.rs#L200

## Solution

Remove generic from `Signature`. Instead, we now convert signature into `k256::Signature` during recovery.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -49,21 +49,13 @@ mod signed;
 pub use signed::{BigIntConversionError, ParseSignedError, Sign, Signed};
 
 mod signature;
-pub use signature::{to_eip155_v, Parity, SignatureError};
+pub use signature::{to_eip155_v, Parity, Signature, SignatureError};
 
 /// Only available for documentation purposes.
 // Without this visible (not `#[doc(hidden)]`) re-export, `rustdoc` will not generate documentation
 // for the `Signature` type alias below.
 #[cfg(feature = "unstable-doc")]
 pub use signature::Signature as PrivateSignature;
-
-/// An ECDSA Signature, consisting of V, R, and S.
-#[cfg(feature = "k256")]
-pub type Signature = signature::Signature<k256::ecdsa::Signature>;
-
-/// An ECDSA Signature, consisting of V, R, and S.
-#[cfg(not(feature = "k256"))]
-pub type Signature = signature::Signature<()>;
 
 pub mod utils;
 pub use utils::{eip191_hash_message, keccak256, Keccak256};
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -51,12 +51,6 @@ pub use signed::{BigIntConversionError, ParseSignedError, Sign, Signed};
 mod signature;
 pub use signature::{to_eip155_v, Parity, Signature, SignatureError};
 
-/// Only available for documentation purposes.
-// Without this visible (not `#[doc(hidden)]`) re-export, `rustdoc` will not generate documentation
-// for the `Signature` type alias below.
-#[cfg(feature = "unstable-doc")]
-pub use signature::Signature as PrivateSignature;
-
 pub mod utils;
 pub use utils::{eip191_hash_message, keccak256, Keccak256};
 
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -83,7 +83,7 @@ impl TryFrom<Signature> for k256::ecdsa::Signature {
     type Error = k256::ecdsa::Error;
 
     fn try_from(value: Signature) -> Result<Self, Self::Error> {
-        Self::from_scalars(value.r.to_be_bytes(), value.s.to_be_bytes())
+        value.to_k256()
     }
 }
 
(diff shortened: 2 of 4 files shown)


crates/primitives/src/signature/sig.rs
```diff
@@ -8,54 +8,34 @@ use crate::{
...
+const SECP256K1N_ORDER: U256 = U256::from_be_bytes(
+    fixed_bytes!("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141").0,
+);
+
 /// An Ethereum ECDSA signature.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct Signature<T> {
-    /// Memoized ecdsa signature (if any)
-    inner: T,
-
```
mattsse, line 16: this makes sense, this wasn't very useful and would also result in a higher mem footprint

mattsse (approved):
I like this a lot, for the exact reasons you described

> Instead, we now convert signature into k256::Signature during recovery.

this makes sense, we then can do different recovery even

crates/primitives/src/lib.rs
```diff
@@ -49,22 +49,14 @@ mod signed;
...
 
 mod signature;
-pub use signature::{to_eip155_v, Parity, SignatureError};
+pub use signature::{to_eip155_v, Parity, Signature, SignatureError};
 
 /// Only available for documentation purposes.
 // Without this visible (not `#[doc(hidden)]`) re-export, `rustdoc` will not generate documentation
 // for the `Signature` type alias below.
 #[cfg(feature = "unstable-doc")]
 pub use signature::Signature as PrivateSignature;
```
DaniPopes, line 58: this is not needed anymore, also in core meta crate

crates/primitives/src/signature/sig.rs
```diff
@@ -8,54 +8,34 @@ use crate::{
 use alloc::vec::Vec;
 use core::str::FromStr;
 
+/// The order of the secp256k1 curve
+const SECP256K1N_ORDER: U256 = U256::from_be_bytes(
```
DaniPopes, line 12: uint! macro

crates/primitives/src/signature/sig.rs
```diff
@@ -64,43 +44,52 @@ impl FromStr for Signature<()> {
...
         Self::from_signature_and_parity(value.0, value.1).unwrap()
     }
 }
 
+#[cfg(feature = "k256")]
+impl TryFrom<Signature> for k256::ecdsa::Signature {
+    type Error = k256::ecdsa::Error;
+
+    fn try_from(value: Signature) -> Result<Self, Self::Error> {
+        Self::from_scalars(value.r.to_be_bytes(), value.s.to_be_bytes())
```
DaniPopes, line 87: recovery becomes more expensive because this is not a free conversion

crates/primitives/src/signature/sig.rs
```diff
@@ -126,47 +114,35 @@ impl Signature<k256::ecdsa::Signature> {
...
-        r: U256,
-        s: U256,
-        parity: T,
-    ) -> Result<Self, SignatureError> {
-        Self::from_scalars_and_parity(r.into(), s.into(), parity)
-    }
-
-    /// Parses a signature from a byte slice, with a v value
-    #[inline]
-    pub fn from_bytes_and_parity<T: TryInto<Parity, Error = E>, E: Into<SignatureError>>(
```
DaniPopes, line 150: why are we not keeping these?

crates/primitives/src/signature/sig.rs
```diff
@@ -271,15 +251,7 @@ impl Signature<()> {
...
         parity: T,
     ) -> Result<Self, SignatureError> {
-        Ok(Self { inner: (), v: parity.try_into().map_err(Into::into)?, r, s })
-    }
-}
-
-impl<S: Copy> Signature<S> {
-    /// Returns the inner ECDSA signature.
-    #[inline]
-    pub const fn into_inner(self) -> S {
```
DaniPopes, line 281: we can keep this as deprecated for cfg k256 and panic on error to avoid breaking change

crates/primitives/src/signature/sig.rs
DaniPopes: worth looking into easier ways to construct the scalars without going through big endian bytes

crates/primitives/src/signature/sig.rs
klkvr: we are, they were duplicated in two impl blocks

https://github.com/alloy-rs/core/pull/719/files#diff-79b82cfeeba5106f10ab06fb251c628206b2d8f648bcd94e6fe01c6091763d1cR239-R249

crates/primitives/src/signature/sig.rs
DaniPopes: https://github.com/alloy-rs/core/pull/719#discussion_r1741064689

crates/primitives/src/signature/sig.rs
klkvr: I don't think k256 Signature allows instantiation without going through bytes

Given that signature fields are not public, should we consider adding this to it:
```rust
#[cfg(feature = "k256")]
signature: Option<k256::ecdsa::Signature>,
```

which will be set to `None` if invalid signature is instantiated, and recovery will just fail if its `None`

mattsse: > I don't think k256 Signature allows instantiation without going through bytes

it doesn't

I think we should be okay with this overhead. this only affects the alloy sig -> k256 signature conversion and most common is decoding this from rlp

and getting rid of the additional footprint is more impactful, at least from reth pov

mattsse (approved):

crates/primitives/src/signature/sig.rs
```diff
@@ -126,47 +113,43 @@ impl Signature<k256::ecdsa::Signature> {
         .unwrap()
     }
 
+    /// Instantiate a new signature from `r`, `s`, and `v` values.
+    pub const fn new(r: U256, s: U256, v: Parity) -> Self {
```
DaniPopes, line 117: I don't think we need const here, I don't want us to lock ourselves

crates/primitives/src/signature/sig.rs
```diff
@@ -64,43 +43,52 @@ impl FromStr for Signature<()> {
...
 #[cfg(feature = "k256")]
-impl From<(k256::ecdsa::Signature, k256::ecdsa::RecoveryId)> for Signature<k256::ecdsa::Signature> {
+impl From<(k256::ecdsa::Signature, k256::ecdsa::RecoveryId)> for Signature {
     fn from(value: (k256::ecdsa::Signature, k256::ecdsa::RecoveryId)) -> Self {
         Self::from_signature_and_parity(value.0, value.1).unwrap()
     }
 }
 
+#[cfg(feature = "k256")]
+impl TryFrom<Signature> for k256::ecdsa::Signature {
```
DaniPopes, line 82: can we have this as a separate method too? `to_k256` maybe

DaniPopes (changes requested):

DaniPopes (approved):
lgtm

Issue #717, opened 2024-08-31
[Bug] Alloy-sol-type-parser blocking compile of alloy crate


CRossel87a:
### Component

sol-type-parser

### What version of Alloy are you on?

0.3.0

### Operating System

Windows

### Describe the bug

   Compiling alloy-sol-type-parser v0.8.0
error[E0277]: the trait bound `CustomError: StdError` is not satisfied
  --> C:\Users\chris\.cargo\registry\src\index.crates.io-6f17d22bba15001f\alloy-sol-type-parser-0.8.0\src\input.rs:39:13 
   |
39 |             winnow::error::ErrMode::from_external_error(input, winnow::error::ErrorKind::Eof, err)
   |             ^^^^^^^^^^^^^^^^^^^^^^ the trait `StdError` is not implemented for `CustomError`, which is required by `ErrMode<_>: FromExternalError<_, _>`

Running latest Rust version

CRossel87a:
Ah, this only happens if you have ethers in the same dependency list

shuhuiluo:
This issue is also bugging me. I see "the trait `core::error::Error` is not implemented for `CustomError`, which is required by `ErrMode<_>: FromExternalError<_, _>`".

Can confirm removing `ethers` solves the problem. Or add
```rust
#[cfg(not(feature = "std"))]
impl core::error::Error for CustomError {}
```
to `alloy-sol-type-parser` `input.rs`.

DaniPopes:
are you using no_std? you need to enable alloy-sol-type-parser/std feature if winnow/std is also enabled

shuhuiluo:
> are you using no_std? you need to enable alloy-sol-type-parser/std feature if winnow/std is also enabled

I'm. But it doesn't compile with or without `std` feature.

DaniPopes:
I dont believe that, the problem is a cfg(std) impl in winnow, you have to enable std in alloy as well for it to work

shuhuiluo:
> I dont believe that, the problem is a cfg(std) impl in winnow, you have to enable std in alloy as well for it to work

@DaniPopes Can you take a look at [this branch](https://github.com/shuhuiluo/uniswap-v3-sdk-rs/tree/sol-type-parser-error)? I can only make it compile by removing `ethers` or `impl core::error::Error for CustomError`.

Wollac:
[...]

PR #720, opened 2024-09-03
fix(sol-type-parser): winnow std error

DaniPopes:
Temporary hotfix as the implementation is correct but errors anyway, see issue for more details.

Fixes https://github.com/alloy-rs/core/issues/717.


mattsse (approved):

kayabaNerve:
FWIW, I just encountered this while upgrading to 0.3 so I'd love a patch release.

PR #727, opened 2024-09-06
fix: `no_std` and workflow

refcell:
### Description

Adds a `no_std` github action workflow.

`alloy-primitives` is breaking `no_std` due to `derive_more` having `full` features.


DaniPopes (approved):

refcell:
I've added this to #728 @DaniPopes

refcell:
Oh oops just saw you fixed this

scripts/check_no_std.sh
```diff
@@ -0,0 +1,26 @@
...
+    alloy-primitives
+    # alloy-sol-macro
+    # alloy-sol-macro-expander
+    # alloy-sol-macro-input
+    alloy-sol-type-parser
+    alloy-sol-types
+    # syn-solidity
+)
+
+cmd=(cargo +stable hack check --no-default-features --target "$target")
```
refcell, line 20: I think we'll either need to use `+nightly` here or bump the rust-version in the cargo toml to `1.81` since the core Error trait is stabilized.

DaniPopes: I can just disable the error feature

mattsse (approved):

refcell:
Thanks for updating quickly @DaniPopes 🙏

DaniPopes:
Released 0.8.2 with this fix. Thanks for the PR(s)!

PR #728, opened 2024-09-06
fix: Turn off derive_more default features

refcell:
### Description

Turns off default-features for the `derive_more` dep to support `no_std`.


mattsse (approved):

refcell:
Fixing... Still need `full` feat flag

refcell:
This now passes the `no_std` check. Can add to this pr to demonstrate

PR #732, opened 2024-09-07
feat: prepare reth Signature migration to alloy

leruaa:
## Motivation

To achieve https://github.com/paradigmxyz/reth/issues/10750 and replace reth Signature type by alloy one, 

## Solution

The following is needed:

- impl Default for Parity
- impl Hash and Default for Signature
- add as_hex_bytes() and size() to Signature

I also moved some unit tests from reth to alloy.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/parity.rs
+++ b/crates/primitives/src/signature/parity.rs
@@ -16,16 +16,16 @@ pub enum Parity {
     Parity(bool),
 }
 
+impl Default for Parity {
+    fn default() -> Self {
+        Self::Parity(false)
+    }
+}
+
 #[cfg(feature = "k256")]
 impl From<k256::ecdsa::RecoveryId> for Parity {
     fn from(value: k256::ecdsa::RecoveryId) -> Self {
         Self::Parity(value.is_y_odd())
-    }
-}
-
-impl Default for Parity {
-    fn default() -> Self {
-        Self::Parity(false)
     }
 }
 
--- a/crates/primitives/src/signature/parity.rs
+++ b/crates/primitives/src/signature/parity.rs
@@ -23,6 +23,12 @@ impl From<k256::ecdsa::RecoveryId> for Parity {
     }
 }
 
+impl Default for Parity {
+    fn default() -> Self {
+        Self::Parity(false)
+    }
+}
+
 impl TryFrom<U64> for Parity {
     type Error = <Self as TryFrom<u64>>::Error;
     fn try_from(value: U64) -> Result<Self, Self::Error> {
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -3,7 +3,7 @@
 use crate::{
     hex,
     signature::{Parity, SignatureError},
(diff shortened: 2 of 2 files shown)


crates/primitives/src/signature/parity.rs
```diff
@@ -23,6 +23,12 @@ impl From<k256::ecdsa::RecoveryId> for Parity {
     }
 }
 
+impl Default for Parity {
```
DaniPopes, line 26: pls move above the from impl above

crates/primitives/src/signature/sig.rs
```diff
@@ -351,6 +356,12 @@ impl Signature {
         self.write_rlp_v(out);
         self.write_rlp_rs(out);
     }
+
+    /// Calculates a heuristic for the in-memory size of the [Signature].
+    #[inline]
+    pub const fn size(&self) -> usize {
```
DaniPopes, line 362: i'd rather not have these methods

DaniPopes (changes requested):

crates/primitives/src/signature/sig.rs
```diff
@@ -314,6 +314,11 @@ impl Signature {
         sig
     }
 
+    /// Turn this signature into its hex-encoded representation.
+    pub fn as_hex_bytes(&self) -> Bytes {
```
DaniPopes, line 318: this too, 1 the name is wrong 2 just use as_bytes, if you need it as Bytes then do .into or Bytes::from

DaniPopes (changes requested):

crates/primitives/src/signature/sig.rs
```diff
@@ -13,7 +13,7 @@ const SECP256K1N_ORDER: U256 =
     uint!(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141_U256);
 
 /// An Ethereum ECDSA signature.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Hash, Default, PartialEq, Eq)]
 pub struct Signature {
```
klkvr, line 17: I don't think we should add this. IMO it's not great in general to have `Default` impl producing invalid type. we already have `Signature::test_signature` which can be used in tests

leruaa: We'll need to remove `Default` on a few type encapsulating `Signature` on reth: 

* BlobTransaction
* TransactionSigned
* TransactionSignedNoHash
* TransactionSignedEcRecovered

I was expecting more than that, seems doable.

klkvr: feel free to keep those as manual impls, I think we only use them for testing, just don't want to get this into core

DaniPopes (approved):

mattsse (approved):

klkvr (approved):

PR #737, opened 2024-09-13
feat: support Keccak with sha3

xJonathanLEI:
Adds a new feature `sha3-keccak` to `alloy-primitives` that when enabled, uses `sha3` for Keccak. This can be desirable as in some cases `sha3` is faster.

This feature is added in a completely backward-compatible way: `tiny-keccak` is still the default, and downstream libs/apps will _not_ suddenly start using `sha3` by just doing `cargo update`.

### DIFF
--- a/crates/primitives/src/utils/mod.rs
+++ b/crates/primitives/src/utils/mod.rs
@@ -257,7 +257,7 @@ impl Keccak256 {
     /// Absorbs additional input. Can be called multiple times.
     #[inline]
     pub fn update(&mut self, bytes: impl AsRef<[u8]>) {
-        keccak256_state::State::update(&mut self.state, bytes.as_ref());
+        self.state.update(bytes.as_ref());
     }
 
     /// Pad and squeeze the state.
--- a/crates/primitives/src/utils/mod.rs
+++ b/crates/primitives/src/utils/mod.rs
@@ -13,16 +13,6 @@ pub use units::{
     format_ether, format_units, parse_ether, parse_units, ParseUnits, Unit, UnitsError,
 };
 
-cfg_if! {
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
Yo any update on this?

onbjerg (approved):
lgtm impl wise, pending either @DaniPopes or @mattsse if we want this in

mattsse (commented):
tbh not super familiar with the diff, but do they even produce the same hashes?

onbjerg:
yeah they do produce the same hashes, both crates are impls of https://csrc.nist.gov/pubs/fips/202/final

crates/primitives/src/utils/mod.rs
```diff
@@ -16,6 +16,8 @@ pub use units::{
 cfg_if! {
     if #[cfg(all(feature = "asm-keccak", not(miri)))] {
         use keccak_asm::Digest as _;
+    } else if #[cfg(all(feature = "sha3-keccak", not(miri)))] {
+        use sha3::Digest as _;
```
mattsse, line 20: what happens if both asm-keccak and sha3.keccak are enabled? does it compile?

mattsse (commented):
not opposed to it, but defer to @DaniPopes 

1 question about activated both feature gated hashers

crates/primitives/src/utils/mod.rs
xJonathanLEI: Yep this is checked on CI I believe:

https://github.com/alloy-rs/core/blob/022c63db87b6bca348dd8e55f86c94a614b99cb6/.github/workflows/ci.yml#L100-L111

DaniPopes (approved):
Thanks! I've simplified the cfgs by consolidating them into one place

PR #742, opened 2024-09-20
Add custom serialization for Address

jenpaff:
## Motivation

Partially closes https://github.com/foundry-rs/foundry/issues/8715
We want to serialize Address type in checksum format. 

## Solution
Solution is to implement a custom serialization function for `Address` type. 

I ran into some issues: 
1) The `impl_serde` macro conflicts with `impl Serialize for Address` (`Address` is wrapped with the `wrap_fixed_bytes` macro which calls `impl_serde`), so I added a workaround in `impl_serde` to bypass `Address` type. Another solution would be to just move the impl directly into the macro- if preferred. 

2) There's another type `EIP712Domain` which uses the type `Address`, I wanted to double check that it is okay for this to also be serialized in checksum format, I added another unit test to make it explicit.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/serde.rs
+++ b/crates/primitives/src/bits/serde.rs
@@ -1,9 +1,30 @@
-use super::FixedBytes;
-use core::fmt;
+use super::{Address, FixedBytes};
+use alloc::string::String;
+use core::{fmt, str::FromStr};
 use serde::{
     de::{self, Visitor},
     Deserialize, Deserializer, Serialize, Serializer,
 };
+
+impl Serialize for Address {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let checksum_address = self.to_checksum(None);
+        serializer.serialize_str(&checksum_address)
+    }
+}
+
+impl<'de> Deserialize<'de> for Address {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        Self::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
 
 impl<const N: usize> Serialize for FixedBytes<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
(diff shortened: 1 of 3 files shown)


zerosnacks:
1. Implementation looks good, I don't think the NOOP behavior in the macro is a problem but documenting it would be good. Perhaps there is a better way of handling this.
2. It doesn't look like EIP712 enforces EIP-55, and nothing specifically relating to the `verifyingContract` field. From preliminary tests it looks like checksumming was already handled here so the resulting hash should stay the same - worth double checking.

cc @DaniPopes please correct me if I'm wrong

Tip: for CI `fmt` task failing run: `cargo +nightly fmt --all --check`

crates/primitives/src/bits/serde.rs
```diff
@@ -1,10 +1,31 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for Address {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        Self::from_str(&s).map_err(serde::de::Error::custom)
```
DaniPopes, line 25:
```suggestion
        Deserialize::deserialize(deserializer).map(Self)
```

crates/primitives/src/bits/serde.rs
```diff
@@ -1,10 +1,31 @@
...
     Deserialize, Deserializer, Serialize, Serializer,
 };
 
+impl Serialize for Address {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let checksum_address = self.to_checksum(None);
+        serializer.serialize_str(&checksum_address)
```
DaniPopes, line 15: Use `to_checksum_buffer` (more efficient), and add the same `is_human_readable` as below in `impl<const N: usize> Serialize for FixedBytes<N>`

DaniPopes (commented):
LGTM, nits

Address override in impl_serde macro is fine, I can't think of a better way to do this

zerosnacks (approved):

DaniPopes (approved):

PR #743, opened 2024-09-20
feat(primitives): implement `map` module

DaniPopes:
Add `mod map` in `alloy-primitives` which contains type aliases for different hashmaps with different hashers.

Contains a set of:
- hashmap, hashset (from hashbrown or std); indexmap, indexset
- default hasher, fxhash
- fixed-bytes hasher for FixedBytes<N> and aliases

Default hasher is: fxhash random seed ("map-fxhash" + "std" + "rand") > fxhash no seed ("map-fxhash") > hashbrown default hasher
(can't have random seed in no_std AFAICT)

New features:
- "map" - enables the module
- "map-fxhash" - switches default hasher to fxhash and re-exports
- "map-hashbrown" - forces HashMap to be a re-export of hashbrown instead of std (default for no_std)
- "map-indexmap" - indexmap re-exports
- "nightly" - passes through "nightly" to dependencies, useful for enabling override of `Hasher::write_length_prefix`


crates/primitives/src/map/fixed.rs
```diff
@@ -0,0 +1,152 @@
...
+        pub type FbIndexSet<const N: usize, S = DefaultHashBuilder> =
+            indexmap::IndexSet<FixedBytes<N>, BuildFixedBytesHasher<N, S>>;
+    }
+}
+
+macro_rules! fb_alias_maps {
+    ($($ty:ident < $n:literal >),* $(,)?) => { paste::paste! {
+        $(
+            #[doc = concat!("[`HashMap`] optimized for hashing [`", stringify!($ty), "`].")]
+            pub type [<$ty HashMap>]<V, S = DefaultHashBuilder> = HashMap<$ty, V, S>;
```
DaniPopes, line 28: S should be hardcoded to `BuilderFixedBytesHasher<$n>`

mattsse (approved):
cool, this should be useful

the stable build error looks odd

DaniPopes:
It's due to "nightly" feature due to --all-features on stable

onbjerg (approved):

rkrasiuk (approved):
great

Issue #744, opened 2024-09-24
[Bug] Solidity type alias yields `unresolved custom type:`


gpsanant:
### Component

sol! macro

### What version of Alloy are you on?

v0.8.3

### Operating System

macOS (Apple Silicon)

### Describe the bug

The following abi 

[DelegationManager.json](https://github.com/user-attachments/files/17109326/DelegationManager.json)

along with the following code

```rust
mod core {
    alloy::sol!(
        #[allow(missing_docs)]
        #[sol(rpc)]
        DelegationManager,
        "../abis/DelegationManager.json"
    );
}
```

```rust
error: unresolved custom type: DelegatedShares
  --> scenario-generator/src/bin/populate_src.rs:63:5
   |
63 | /     alloy::sol!(
64 | |         #[allow(missing_docs)]
65 | |         #[sol(rpc)]
66 | |         DelegationManager,
67 | |         "../abis/DelegationManager.json"
68 | |     );
   | |_____^
   |
   = note: this error originates in the macro `$crate::sol_types::sol` which comes from the expansion of the macro `alloy::sol` (in Nightly builds, run with -Z macro-backtrace for more info)
```

If you prefer to build the contracts yourself:
```
git clone git@github.com:Layr-Labs/eigenlayer-contracts.git
git checkout slashing-magnitues
forge b -C src/contracts
```

See `out/DelegationManager.sol/DelegationManager.json`.

I'm using a type alias `DelegatedShares` for `uint256`.

mattsse:
yeah this looks like an issue with type aliases

```json
         {
            "name": "delegatedShares",
            "type": "uint256[]",
            "internalType": "DelegatedShares[]"
          }
```

DaniPopes:
There's 2 issues here:
1. We're ignoring UDVT arrays, fixed in https://github.com/alloy-rs/core/pull/745
2. After this fix, it still fails because the type doesn't have an assigned contract because it is declared globally, this requires a bit more work to fix

gpsanant:
Thanks for the prompt response.

For now, I'm actually not using the functions with the custom types so I'll just redefine interfaces.

If I included the types inside of the library contract in which they're defined, would that address (2)?

DaniPopes:
Yes I think so, if you were to move the `type DelegatedShares is uint256;` inside of one of the interfaces

gpsanant:
Cool, confirming that
1. getting rid of the alias array
2. stuffing the alias definition inside of the library

[...]

PR #745, opened 2024-09-24
fix(json-abi): correct to-sol for UDVT arrays in structs

DaniPopes:
Ref https://github.com/alloy-rs/core/issues/744

### DIFF
--- a/crates/json-abi/src/to_sol.rs
+++ b/crates/json-abi/src/to_sol.rs
@@ -229,9 +229,11 @@ impl<'a> InternalTypes<'a> {
                 }
             }
             Some(it @ InternalType::Other { contract, ty }) => {
-                // `Other` is a UDVT if it's not a basic Solidity type and not an array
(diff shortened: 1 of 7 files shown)


crates/json-abi/src/to_sol.rs
```diff
@@ -229,9 +229,11 @@ impl<'a> InternalTypes<'a> {
                 }
             }
             Some(it @ InternalType::Other { contract, ty }) => {
-                // `Other` is a UDVT if it's not a basic Solidity type and not an array
+                // `Other` is a UDVT if it's not a basic Solidity type.
                 if let Some(it) = it.other_specifier() {
-                    if it.try_basic_solidity().is_err() && !it.is_array() {
+                    if it.try_basic_solidity().is_err() {
+                        let ty = ty.split('[').next().unwrap();
+                        let real_ty = real_ty.split('[').next().unwrap();
```
mattsse, line 236: this should work for non-arrays and arrays alike

mattsse (approved):

PR #755, opened 2024-10-04
feat: add `Default` for `Sealed<T>`

emhane:
## Motivation

`Default` is required trait bound for reth integration of `Sealed<T>`. Ref https://github.com/paradigmxyz/reth/issues/11123.

## Solution

Derives `Default` for `Sealed<T>`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -4,7 +4,7 @@ use crate::B256;
 ///
 /// We do not implement any specific hashing algorithm here. Instead types
 /// implement the [`Sealable`] trait to provide define their own hash.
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this is also problematic, we can only safely do this if we seal in the manual default impl

emhane:
> this is also problematic, we can only safely do this if we seal in the manual default impl

same here, pls open issue in reth to manual impl `Default` for all types that use `SealedBlock`

mattsse (approved):

PR #762, opened 2024-10-04
Derive `Arbitrary` for `Sealed<T>`

emhane:
## Motivation

`Arbitrary` is required trait bound for reth integration of `Sealed<T>`. Ref https://github.com/paradigmxyz/reth/issues/11123.

## Solution

Derives `Arbitrary` for `Sealed<T>`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -6,6 +6,7 @@ use crate::B256;
 /// implement the [`Sealable`] trait to provide define their own hash.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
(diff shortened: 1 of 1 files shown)


crates/primitives/src/sealed.rs
```diff
@@ -6,6 +6,7 @@ use crate::B256;
 /// implement the [`Sealable`] trait to provide define their own hash.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(derive_arbitrary::Arbitrary, proptest_derive::Arbitrary))]
```
mattsse, line 9: this could be problematic because this generates an arbitrary hash,
we should probably impl for T: Sealable and then seal?

mattsse (changes requested):

mattsse (approved):

mattsse (approved):

PR #768, opened 2024-10-10
feat: add logs_bloom

mattsse:
copied from

https://github.com/paradigmxyz/reth/blob/33d5eb322d32134c0875219f6e562824fbeca87e/crates/primitives-traits/src/log.rs#L4-L14

this only uses primitive types and can live here

ref https://github.com/alloy-rs/op-alloy/pull/165#pullrequestreview-2358347227

### DIFF
--- a/crates/primitives/src/log/mod.rs
+++ b/crates/primitives/src/log/mod.rs
@@ -1,8 +1,20 @@
-use crate::{Address, Bytes, B256};
+use crate::{Address, Bloom, Bytes, B256};
 use alloc::vec::Vec;
 
 #[cfg(feature = "serde")]
 mod serde;
+
+/// Compute the logs bloom filter for the given logs.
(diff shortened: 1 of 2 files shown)


crates/primitives/src/log/mod.rs
```diff
@@ -1,9 +1,21 @@
...
 mod serde;
 
+/// Compute the logs bloom filter for the given logs.
+pub fn logs_bloom<'a>(logs: impl IntoIterator<Item = &'a Log>) -> Bloom {
+    let mut bloom = Bloom::ZERO;
+    for log in logs {
+        bloom.m3_2048(log.address.as_slice());
+        for topic in log.topics() {
+            bloom.m3_2048(topic.as_slice());
+        }
```
DaniPopes, line 14: `bloom.accrue_log`

DaniPopes (changes requested):
Needs clippy

DaniPopes (approved):

PR #770, opened 2024-10-10
fix(alloy-sol-macro): allow clippy::pub_underscore_fields on `sol!` output

### LINKED ISSUE #769: [Bug] `sol!` should emit `#[allow(clippy::pub_underscore_fields)]`
### Component

sol! macro

alloy v0.4.2

### Operating System

None

### Describe the bug

since `sol!` doesn't change the case of fields, it should allow `clippy::pub_underscore_fields` (and perhaps allow some other lints as well?)

```rust
sol! {
    function someFunction(
        bytes calldata _ignored
    );
}
```

```
warning: field marked as public but also inferred as unused because it's prefixed with `_`
  --> src/main.rs:53:1
   |
53 | / alloy::sol! {
54 | |     function someFunction(
55 | |         bytes calldata _ignored
56 | |     );
57 | | }
   | |_^
   |
   = help: consider removing the underscore, or making the field private
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#pub_underscore_fields
note: the lint level is defined here
  --> src/main.rs:1:9
   |
1  | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::pub_underscore_fields)]` implied by `#[warn(clippy::pedantic)]`
   = note: this warning originates in the macro `$crate::sol_types::sol` which comes from the expansion of the macro `alloy::sol` (in Nightly builds, run with -Z macro-backtrace for more info)
```

and yes, this can be fixed by putting `#[allow(clippy::pub_underscore_fields)]` everywhere where this is an issue, but that's quite annoying when there are large contracts being copied into the `sol!` macro

benluelo:
fixes #769, not sure if tests are required here - if they are, just let me know how I should add a test for this

### DIFF
--- a/crates/sol-macro-expander/src/expand/contract.rs
+++ b/crates/sol-macro-expander/src/expand/contract.rs
@@ -488,7 +488,7 @@ pub(super) fn expand(cx: &mut ExpCtxt<'_>, contract: &ItemContract) -> Result<To
         #mod_descr_doc
         #(#mod_attrs)*
         #mod_iface_doc
-        #[allow(non_camel_case_types, non_snake_case, clippy::style)]
+        #[allow(non_camel_case_types, non_snake_case, clippy::pub_underscore_fields, clippy::style)]
         pub mod #name {
             use super::*;
             use #alloy_sol_types as alloy_sol_types;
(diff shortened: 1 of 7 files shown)


DaniPopes (approved):
Thanks!

Issue #767, opened 2024-10-09
Default hasher `foldhash` queries the time


Wollac:
### Component

primitives

### What version of Alloy are you on?

alloy-primitives v0.8.6

### Operating System

None

### Describe the bug

The recent update to `foldhash` as the default hasher for map in `v0.8.6` introduces a dependency on `std::time` within `foldhash`. This poses a problem for targets like `zkvm`, where time is not available, leading to errors.

This change effectively breaks compatibility with `zkvm` and potentially other targets lacking `std::time`.  Projects relying on Alloy and targeting these platforms will no longer work successfully.

DaniPopes:
I've removed the default in https://github.com/alloy-rs/core/pull/771, but this is also the default hasher for Hashbrown, so consider opening an issue/PR in foldhash to extend disabling the time query cfg

PR #771, opened 2024-10-11
fix: don't enable foldhash by default

DaniPopes:
Fixes https://github.com/alloy-rs/core/issues/767

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -95,7 +95,7 @@ criterion.workspace = true
 serde_json.workspace = true
 
 [features]
-default = ["std", "map-foldhash"]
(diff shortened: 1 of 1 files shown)


mattsse (approved):

hai-rise:
@DaniPopes @mattsse  I did a naive dep bump for `pevm` today, it was consistently red with up to 35% regression.
https://github.com/risechain/pevm/pull/386/files

image

10-35% regression on a casual `alloy-primitives` bump is very scary, and I don't think most library users have rigid benchmarks to catch it. I'd suggest either:
- Tell people loud and clear to enable `map-foldhash` for performance-critical usage, or
- Revert this and ask users affected by #767 to use something like `no-default`?

PR #773, opened 2024-10-16
feat: expand Seal api

prestwich:
## Motivation

Was working on https://github.com/alloy-rs/alloy/pull/1485 and it seemed like there were a few holes in the API

this is a breaking change as it flattens the serde object

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -131,13 +131,18 @@ pub trait Sealable: Sized {
         Sealed::new(self)
     }
 
-    /// Seal the object by calculating the hash. This may be slow.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense, this type should flatten and renaming seal -> hash is more useful

pending @DaniPopes

prestwich:
> makes sense, this type should flatten and renaming seal -> hash is more useful

we can rename to hash if you'd prefer. right now it only allows that as a deser alias, while still producing ser by default

This PR includes a serde helper mod for renaming to hash
https://github.com/alloy-rs/alloy/pull/1485

DaniPopes (approved):

PR #774, opened 2024-10-17
feat: add helper to eip155 variant

mattsse:
sometimes we need special checks for 155 or need to know if it is 155

adds two helpers for this

ref https://github.com/bluealloy/revm/pull/1824/files

### DIFF
--- a/crates/primitives/src/signature/parity.rs
+++ b/crates/primitives/src/signature/parity.rs
@@ -76,6 +76,25 @@ impl Parity {
         }
     }
 
+    /// Applies a check to the [`Parity::Eip155`] v value if this instance is a [`Parity::Eip155`].
+    ///
+    /// Returns [`Ok`] if the instance is not [`Parity::Eip155`]
+    pub fn ensure_eip155<F, E>(&self, f: F) -> Result<(), E>
+    where
+        F: FnOnce(u64) -> Result<(), E>,
+    {
+        if let Self::Eip155(v) = self {
+            f(*v)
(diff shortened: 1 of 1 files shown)


crates/primitives/src/signature/parity.rs
```diff
@@ -76,6 +76,25 @@ impl Parity {
         }
     }
 
+    /// Applies a check to the [`Parity::Eip155`] v value if this instance is a [`Parity::Eip155`].
+    ///
+    /// Returns [`Ok`] if the instance is not [`Parity::Eip155`]
+    pub fn ensure_eip155<F, E>(&self, f: F) -> Result<(), E>
```
DaniPopes, line 82: This is too specific, doesn't make sense to have as a method, `is_` and `as_` are enough

mattsse: hmm this would make things like:


```
            // Check y_parity
            if let Parity::Eip155(parity) = auth.signature().v() {
                if parity > u8::MAX as u64 {
                    return Err(InvalidAuthorization::InvalidYParity);
                }
            }
            Ok(())
```

slightly easier imo

DaniPopes: the same is achievable with as_eip().is_some_and(), again this is too specific

mattsse: still think this is nice to have and doesn't hurt really, but closing because can be solved with some extra calls as well

Issue #705, opened 2024-08-05
[Bug] Signatures from sign_hash return Parity, but if converting to Vec and restoring, it returns NonEip155


gusinacio:
### Component

primitives

### What version of Alloy are you on?

0.2.1

### Operating System

None

### Describe the bug

The following code doesn't work, it looks like it's returned Parity() every signature, but if we convert to bytes and restore elsewhere, it returns NonEip155.

The following test fails:
```rust
    #[test]
    fn test_parity() {
        let signer = PrivateKeySigner::random();
        // The message to sign.
        let message = b"hello";

        // Sign the message asynchronously with the signer.
        let signature = signer.sign_message_sync(message).unwrap();

        let value = signature.as_bytes().to_vec();
        let recovered_signature: Signature = value.as_slice().try_into().unwrap();
        assert_eq!(signature, recovered_signature);
    }

```

The error is the following:
```
assertion `left == right` failed
  left: Signature { inner: ecdsa::Signature<Secp256k1>(33828FC1FE49FB3177AA80459AE130BD8147158000FA701D7E33BC33B8EDDF25586239AF447392DD8AC6DACBE5E3F91B72AB5593BE3ED6DC702489CA4E2BE1B6), v: Parity(false), r: 23298637575944829774526378124809190621347391490807772432023406486951885135653, s: 39977079812416937747731152449858872909720600328418637835900651766296202699190 }
 right: Signature { inner: ecdsa::Signature<Secp256k1>(33828FC1FE49FB3177AA80459AE130BD8147158000FA701D7E33BC33B8EDDF25586239AF447392DD8AC6DACBE5E3F91B72AB5593BE3ED6DC702489CA4E2BE1B6), v: NonEip155(false), r: 23298637575944829774526378124809190621347391490807772432023406486951885135653, s: 39977079812416937747731152449858872909720600328418637835900651766296202699190 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test tap::context::rav::test::test_parity ... FAILED

failures:

failures:
    test::test_parity
```

mattsse:
hmm these are indeed not symmetrical

https://github.com/alloy-rs/core/blob/adcf7adfa1f35c56e6331bab85b8c56d32a465f1/crates/primitives/src/signature/parity.rs#L19-L24

[...]

PR #776, opened 2024-10-20
refactor: use simple boolean for parity in signature

klkvr:
## Motivation

Right now we use `Parity` enum for parity value in signature: https://github.com/alloy-rs/core/blob/9c53778e4d40a8ac5b7d3e36221efc1d79fec050/crates/primitives/src/signature/parity.rs#L10-L17

Because of it containing EIP-155 related variants it basically captures transaction-related logic which does not belong to primitives and results in inconsistencies and need for dynamic checks. e.g https://github.com/alloy-rs/alloy/pull/1305 https://github.com/alloy-rs/core/issues/705 https://github.com/alloy-rs/alloy/issues/1510

Primitive signature representation in context of Ethereum is `(U256,U256,bool)`, this PR changes `Signature` to have this structure and leaving specialized parity encoding to types holding signatures and having enough context to perform correct encoding.

At this point in any context we know how parity should be encoded and storing a simple boolean is enough to figure this out. i.e legacy transaction knows its chain_id and can adjust parity accordingly.

API changes in this PR:
- serde now defaults to serializing/deserializing parity as `"yParity": "0x{0,1}"`
- rlp now defaults to encoding/decoding parity as boolean
- `write_rlp_vrs` now accepts encodable `v` value as a parameter, allowing callers to provide arbitrary encoding for parity
- `decode_rlp_vrs` now accepts a `decode_parity` closure allowing callers to provide arbitrary decoding for parity, possibly capturing additional context (e.g legacy chain_id)

This PR is accompanied with PR in alloy to demonstrate/discuss needed API changes https://github.com/alloy-rs/alloy/pull/1540 https://github.com/alloy-rs/eips/pull/12

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -54,7 +54,9 @@ mod signed;
 pub use signed::{BigIntConversionError, ParseSignedError, Sign, Signed};
 
 mod signature;
-pub use signature::{to_eip155_v, Parity, Signature, SignatureError};
+pub use signature::{
+    normalize_v, to_eip155_v, Parity, PrimitiveSignature, Signature, SignatureError,
+};
 
 pub mod utils;
 pub use utils::{eip191_hash_message, keccak256, Keccak256};
--- a/crates/primitives/src/signature/mod.rs
+++ b/crates/primitives/src/signature/mod.rs
@@ -8,4 +8,7 @@ mod sig;
 pub use sig::Signature;
 
 mod utils;
-pub use utils::to_eip155_v;
+pub use utils::{normalize_v, to_eip155_v};
+
+mod primitive_sig;
+pub use primitive_sig::PrimitiveSignature;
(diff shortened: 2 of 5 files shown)


crates/primitives/src/signature/sig.rs
```diff
@@ -90,15 +88,17 @@ impl TryFrom<Signature> for k256::ecdsa::Signature {
 #[cfg(feature = "rlp")]
 impl Signature {
     /// Decode an RLP-encoded VRS signature.
-    pub fn decode_rlp_vrs(buf: &mut &[u8]) -> Result<Self, alloy_rlp::Error> {
+    pub fn decode_rlp_vrs(
```
mattsse, line 91: this needs more context on the parity decoding closure

mattsse (commented):
I personally much prefer this over the enum approach.
another option would be just an integer but given that we need parity as bool, this makes more sense.

I wonder if we should remove the rlp trait impls for this entirely to avoid footguns, because en/decoding is tx specific?

regarding rollout, this will require a few changes downstream, so I wonder if we should instead introduce this a new type first, transition everything, then remove the old Signature type?

klkvr:
Extracted new signature type to `primitive_sig::PrimitiveSignature`

mattsse (approved):
lgtm

PR #780, opened 2024-10-23
chore: bump MSRV to 1.81 & use `core::error::Error` in place of `std`

Evalir:
## Motivation

Bumping the MSRV to 1.81 lets us use `core::error::Error` as it was finally added to the core library.

The only places where a `std` feature gate for errors was not removed, was because `const-hex` and/or `ruint` would need to do this too. @DaniPopes wdyt?

## Solution

- Bump the MSRV to 1.81
- Remove `std` feature gates for error impls, use `core::error::Error`
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
Makes sense, thanks!

DaniPopes:
Reverted, waiting to re-apply like in https://github.com/alloy-rs/alloy/pull/1552

Evalir:
sounds good—1.83 should be out in exactly a month so we can include it then

PR #790, opened 2024-10-28
chore: bump MSRV to 1.81

DaniPopes:
Reverts https://github.com/alloy-rs/core/pull/789, reapplying #780 and #781.


michaelsproul:
Just want to confirm, but your MSRV policy allows the MSRV to change for a patch release?

I think this is a matter of taste, but I was a bit surprised by this, because it means a `cargo update` can necessitate bumping the MSRV (which is the case for Lighthouse updating alloy right now). I can see why this is just simpler -- and we can definitely make it work -- just wanted to let you know that this is something we noticed.

DaniPopes:
Yes that is right, you can check out the policy in the READMEs of each repository.

PR #807, opened 2024-11-17
Fix broken link to abi module documentation in README.md

harpocryptes:
That link gives a 404 error, at least when viewed on github. I suppose this is the intended target.


DaniPopes:
Thanks however all links in the crates' README are catered towards docs.rs, we unfortunately most of the time can't have both

PR #810, opened 2024-11-25
Make Signature::new a const fn

kpp:
## Motivation

Before https://github.com/alloy-rs/core/pull/719 Signature fields were public and it was possible to create `Signature { r, s, odd_y_parity }` in a const context.

## Solution

Make `Signature::new` a const fn.

## PR Checklist

- [x] ~~Added Tests~~ no need.
- [x] ~~Added Documentation~~ no need.
- [x] ~~Breaking changes~~ no need.

### DIFF
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -114,8 +114,7 @@ impl Signature {
     }
 
     /// Instantiate a new signature from `r`, `s`, and `v` values.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
@klkvr I believe before we change the sig this was non const 

but we can make this const now I believe?

DaniPopes (approved):

klkvr (approved):

PR #818, opened 2024-12-04
Remove unsafe code from macro expansions

nbaksalyar:
## Motivation

- The `sol!` macro cannot be used in projects with strict unsafe code policies.
- The current unsafe implementation of `abi_decode_raw` is not efficient: binary search (`O(log n)`) is slower than a simple jump table (`O(1)` or `O(n)` at worst, for e.g. [compare the MIR/asm output](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0b66ad61f00025c8e9c1e4a1a93a2ab2)).

## Solution

- Switch `abi_decode_raw` to use a constant jump table instead of a runtime binary search.
- Generate safe code to convert between enum variants and `u8` statically.
- Make the `as_u8` function constant.
- Additionally, allow `clippy::empty_structs_with_brackets` in the generated code.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro-expander/src/expand/contract.rs
+++ b/crates/sol-macro-expander/src/expand/contract.rs
@@ -488,7 +488,7 @@ pub(super) fn expand(cx: &mut ExpCtxt<'_>, contract: &ItemContract) -> Result<To
         #mod_descr_doc
         #(#mod_attrs)*
         #mod_iface_doc
-        #[allow(non_camel_case_types, non_snake_case, clippy::pub_underscore_fields, clippy::style)]
+        #[allow(non_camel_case_types, non_snake_case, clippy::pub_underscore_fields, clippy::style, clippy::empty_structs_with_brackets)]
         pub mod #name {
             use super::*;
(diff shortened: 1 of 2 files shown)


DaniPopes:
> The current unsafe implementation of abi_decode_raw is not efficient: binary search (O(log n)) is slower than a simple jump table (O(1) or O(n) at worst, for e.g. compare the MIR/asm output).

This is not true, I would recommend you compare the LLVM IR and/or assembly output for a large calls enum, for example [Foundry's `Vm`](https://github.com/foundry-rs/foundry/blob/00efa0d5965269149f374ba142fb1c3c7edd6c94/crates/cheatcodes/spec/src/vm.rs#L11). The binary search gets unrolled and made branchless at compile time, while the switch gets codegen'd as a binary search regardless because all the match arms have different code sizes.

The other changes are fine.

DaniPopes (approved):

nbaksalyar:
Thank you!

PR #819, opened 2024-12-05
chore: mark `Signature` as deprecated

moricho:
## Motivation

It might be better to notify users in advance before completely migrating from Signature to PrimitiveSignature.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/mod.rs
+++ b/crates/primitives/src/signature/mod.rs
@@ -1,10 +1,14 @@
 mod error;
 pub use error::SignatureError;
 
+#[allow(deprecated)]
 mod parity;
+#[allow(deprecated)]
 pub use parity::Parity;
 
+#[allow(deprecated)]
 mod sig;
+#[allow(deprecated)]
 pub use sig::Signature;
 
 mod utils;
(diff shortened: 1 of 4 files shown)


crates/primitives/src/signature/mod.rs
```diff
@@ -2,9 +2,11 @@ mod error;
 pub use error::SignatureError;
 
 mod parity;
+#[deprecated(since = "0.8.15", note = "see https://github.com/alloy-rs/core/pull/776")]
 pub use parity::Parity;
 
 mod sig;
+#[deprecated(since = "0.8.15", note = "use PrimitiveSignature instead")]
```
mattsse, line 9: could we put those on the struct defs instead?

mattsse (changes requested):

crates/primitives/src/signature/mod.rs
moricho: Fixed it and added some `#[allow(deprecated)]`

DaniPopes (approved):

mattsse (approved):

PR #827, opened 2024-12-19
feat: re-export `rayon` feature

lightsing:
re-export `hashbrown` and `indexmap` `rayon` feature


mattsse (approved):
makes sense to me

pending @DaniPopes

DaniPopes (approved):
Thanks, can you also add it to the alloy-core meta rate?

lightsing:
> Thanks, can you also add it to the alloy-core meta crate?

updated.

Issue #831, opened 2024-12-24
[Bug] PrimitiveSignature doesn't support serialized Signature


gusinacio:
### Component

primitives

### What version of Alloy are you on?

0.8

### Operating System

None

### Describe the bug

My current system uses serialization and deserialization to send signatures via the network. We were using `Signature` but since version 0.4.2, it was deprecated to use `PrimitiveSignature`. The problem is that I don't control the clients and if they update to the latest version or not, so I'm still receiving older signatures.

I had to revert because the deserialization couldn't process the parity if it were `Eip155` or `NonEip155`.
```
app 2024-12-19T18:29:17.724484Z DEBUG jsonrpsee_core::proc_macros_support: Error parsing "receipts" as "Vec < EIP712SignedMessage < Receipt > >": ErrorObject { code: InvalidParams, message: "Invalid params", data: Some(RawValue("invalid y_parity at line 1 column 331" │
│ )) }
```

I don't know if this is considered a breaking change or if you have any suggestions for fixing this problem.

ethsman:
Hello @gusinacio , sorry for leaving a comment here without respecting the team rules. As a blockchain developer, I think my tech stack is fit for the team project. So I want to prove my skills by solving some team issues and join the team as a full-time developer. Could you give me a test task?

mattsse:
@klkvr I assume this happens because of

https://github.com/alloy-rs/core/blob/f8d55dde2109a6c67c9d2004af443a44e6b7a3b6/crates/primitives/src/signature/primitive_sig.rs#L380-L381

gusinacio:
I think it's related to https://github.com/alloy-rs/core/blob/f8d55dde2109a6c67c9d2004af443a44e6b7a3b6/crates/primitives/src/signature/primitive_sig.rs#L389-L390

Since older signatures could have 27/28 for NonEip155 or bigger equal than 35 for Eip155, It's not possible to deserialize using PrimitiveSignature.

klkvr:
@gusinacio with https://github.com/alloy-rs/core/pull/832 this should deserialize correctly, however note that if we receive a signature with eip155 value for `v`, the data about chain id will be lost during deserialization unless the signature is serialized along with chainId key (which is what's happening for legacy transactions for example)

gusinacio:
Thanks for the quick response and resolution! 

I'm using signatures with Eip712 so I don't need information about ChainId because it's already in the domain. As long as it can deserialize correctly and recover the signer address, I'm happy.

[...]

PR #832, opened 2024-12-26
fix: allow non-boolean v values for PrimitiveSignature

klkvr:
## Motivation

Closes https://github.com/alloy-rs/core/issues/831

This still does not allow non-boolean values for `yParity` which should be fine I believe

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):

mattsse:
> This still does not allow non-boolean values for yParity which should be fine I believe

unless the yParity field is supposed to be y-Parity but still allows values > 1 -.-

klkvr:
yeah but hopefully this can only happen for 7702 authorizations

PR #833, opened 2024-12-27
fix: re-enable foldhash on zkvm

Wollac:
As of [v.0.1.4](https://github.com/orlp/foldhash/compare/v0.1.3...v0.1.4), the `foldhash` dependency has fixed the issue on `zkvm` targets originally reported in https://github.com/alloy-rs/core/issues/767. As such, the implemented workaround is no longer necessary as of this release and has been removed in this PR.


DaniPopes (approved):
Thanks!

Issue #816, opened 2024-12-02
[Feature] Support parsing numbers with scientific notation


DaniPopes:
### Component

dyn-abi

### Describe the feature you would like

In coerce.rs https://github.com/alloy-rs/core/blob/main/crates/dyn-abi/src/coerce.rs we should add support for optionally parsing `e<decimal number>` at the end of a number.

From https://github.com/foundry-rs/foundry/issues/8566

nadtech-hub:
:raised_hand_with_fingers_splayed:

PR #835, opened 2024-12-31
feat(dyn-abi): support parse scientific number

9547:
## Motivation

implement and close https://github.com/alloy-rs/core/issues/816

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/dyn-abi/src/coerce.rs
+++ b/crates/dyn-abi/src/coerce.rs
@@ -6,7 +6,7 @@ use core::fmt;
 use hex::FromHexError;
 use parser::{
     new_input,
-    utils::{array_parser, char_parser},
+    utils::{array_parser, char_parser, spanned},
     Input,
 };
 use winnow::{
@@ -345,7 +345,8 @@ fn uint<'i>(len: usize) -> impl Parser<Input<'i>, U256, ContextError> {
             ))
             .parse_next(input)?;
 
-        let intpart = U256::from_str_radix(intpart, 10)
+        let intpart = intpart
+            .parse::<U256>()
             .map_err(|e| ErrMode::from_external_error(input, ErrorKind::Verify, e))?;
         let e = opt(scientific_notation).parse_next(input)?.unwrap_or(0);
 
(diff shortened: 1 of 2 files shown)


crates/dyn-abi/src/coerce.rs
```diff
@@ -422,16 +422,49 @@ fn prefixed_int<'i>(input: &mut Input<'i>) -> PResult<&'i str> {
...
+            scientific_notation,
+        )),
     )
     .parse_next(input)
 }
 
+#[inline]
+fn scientific_notation(input: &mut Input<'_>) -> PResult<usize> {
+    // Check if we have 'e' or 'E' followed by an optional sign and digits
+    if let Some(c) = input.chars().next() {
```
mattsse, line 441: you can use let Some else return err notation here with next + filter

mattsse (changes requested):
style suggestion,

not familiar with the winnow api

crates/dyn-abi/src/error.rs
```diff
@@ -68,6 +68,8 @@ pub enum Error {
     TypeParser(TypeParserError),
     /// [`alloy_sol_types`] error.
     SolTypes(SolTypesError),
+    /// Negative exponent in fixed-point number.
+    NegativeExponent(usize),
```
DaniPopes, line 72: Should move this to the error in coerce.rs

crates/dyn-abi/src/coerce.rs
```diff
@@ -422,16 +424,44 @@ fn prefixed_int<'i>(input: &mut Input<'i>) -> PResult<&'i str> {
...
+    let sign = int_sign(input)?;
+
+    // Parse digits
+    let exp = digit1
+        .parse_next(input)?
+        .parse::<usize>()
+        .map_err(|e| ErrMode::from_external_error(input, ErrorKind::Verify, e))?;
+
+    if sign.is_negative() {
+        Err(ErrMode::from_external_error(input, ErrorKind::Verify, Error::NegativeExponent(exp)))
```
mattsse, line 459: this can be moved below L450

mattsse (changes requested):

crates/dyn-abi/src/coerce.rs
9547: sounds good, thanks

DaniPopes (approved):
Thanks!

PR #836, opened 2025-01-02
feat: re-export rayon traits implementations

lightsing:
re-export rayon traits implementations depends on which HashMap implementation is used


DaniPopes (approved):
Thanks!

Issue #764, opened 2024-10-08
[Feature] Replace `hex-literal` crate with `hexlit` in order to support `"0x"` prefix in the `address!()` macro


gsnoff:
### Component

primitives

### Describe the feature you would like

The `primitives` library defines the `fixed_bytes_macros! {}` internal utility macro, which in turn is used to define exported macros such as `address!()` for compile-time values. The latter takes in a hex-encoded string literal, which is then represented as an `Address` value.

Currently the [`hex-literal`](https://crates.io/crates/hex-literal) crate is responsible for translating hex-encoded string literals into byte arrays in compile time. It is not feature-complete though, notably it lacks the support for `"0x"` prefix.

There exists an alternative [`hexlit`](https://crates.io/crates/hexlit) crate which seems to be a more recent drop-in replacement for `hex-literal`, and which _does_ support additional features such as `"0x"` prefix, as well as some extra kinds of formatting characters, such as hyphen `'-'` and underscore `'_'`.

PR #841, opened 2025-01-03
feat: support 0x in hex! and similar macros

DaniPopes:
Vendor hex-literal and add support for 0x prefixes.

Closes https://github.com/alloy-rs/core/issues/764.


mattsse (approved):
nice!
big fan

PR #848, opened 2025-01-04
fix(primitives): hex macro re-export

DaniPopes:
Unintentional breaking change.


crates/primitives/src/hex_literal.rs
```diff
@@ -94,6 +94,8 @@ macro_rules! hex {
         $crate::hex_literal::decode::<{ $crate::hex_literal::len(STRINGS) }>(STRINGS)
     }};
 }
+#[doc(hidden)] // Use `crate::hex` directly instead!
+pub use crate::hex;
```
DaniPopes, line 98: unfortunately this will also re-export the `hex` crate through this path, so I added `doc(hidden)` to limit the impact

PR #850, opened 2025-01-05
docs: enable some useful rustdoc features on docs.rs

DaniPopes:
- [--generate-link-to-definition](https://doc.rust-lang.org/rustdoc/unstable-features.html#--generate-link-to-definition-generate-links-on-types-in-source-code)
- [--show-type-layout](https://doc.rust-lang.org/rustdoc/unstable-features.html#--show-type-layout-add-a-section-to-each-types-docs-describing-its-memory-layout)
- `--cfg docsrs` [is guaranteed to be enabled](https://docs.rs/about/builds#detecting-docsrs)
- `workspace.metadata` is not read

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,9 +156,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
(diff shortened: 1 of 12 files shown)


mattsse (approved):
TIL

PR #852, opened 2025-01-09
feat(`sol!`): make `#[sol(rpc)]` default if `rpc` feat enabled

yash-atreya:
## Motivation

Currently when expanding contracts in the `sol!` the `rpc` attribute is not enabled by default. Users must explicitly enable using `#[sol(rpc)]`. 

## Solution

Introduces `"rpc"` feature that is propagated from `alloy-core` to `alloy-sol-macro-expander`. 
Enabling this feature expands the contract *with* RPC-related code. 

Now, one can make use of `ContractInstance::new(addr, provider)` without having to specify `#[sol(rpc)]`. 

RPC-related methods can be disabled using `#[sol(rpc = false)]`. 

Note: Enabling this by default requires the user to install `alloy-contract`. 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


gakonst (commented):
If the user has the `contract` feature turned off, or if they haven't installed `alloy-contract`, let's error / warning to prompt the user to do so

DaniPopes (changes requested):
Sorry but I disagree with this, just like with adding derives by default, this makes compilation a lot slower and it's counterintuitive to how the rest of the ecosystem works: derives are opt-in for extra functionality

The default codegen requires the minimum dependencies. RPC codegen is straight up not possible in many contexts like no_std, so this is also a breaking change.

PR #856, opened 2025-01-15
feat(json-abi): add Param.name() accessor

ozgrakkurt:
Adds `Param.name()` for accessing the parameter name.

There is already `Param.name` field but it is hidden in docs.


DaniPopes (approved):
Thanks!

PR #865, opened 2025-02-03
fix(sol-macro)!: rm fake transport from contract expansion

yash-atreya:
## Motivation

After https://github.com/alloy-rs/alloy/pull/1859, we have been using fake traits to avoid breaking `sol!` codegen.
See: https://github.com/alloy-rs/alloy/blob/370dbef8d900ab0e935270b6cc516cdad3c16b7a/crates/contract/src/lib.rs#L43

```rust
    // alloy_contract/lib.rs
    // Fake traits to mitigate `sol!` macro breaking changes.
    pub trait Provider<T, N: Network>: alloy_provider::Provider<N> {}
    impl<N: Network, P: alloy_provider::Provider<N>> Provider<(), N> for P {}

    // This is done so that the compiler can infer the `T` type to be `()`, which is the only type
    // that implements this fake `Transport` trait.
    pub trait Transport {}
    impl Transport for () {}
```

## Solution

This PR makes the necessary changes to the `sol!` contract expansion to avoid using these fake traits.

Companion PR in alloy: https://github.com/alloy-rs/alloy/pull/1996

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


mattsse (approved):
lgtm, but we should probably only merge this once more people are on alloy 0.11 @DaniPopes ?

DaniPopes:
This will require another breaking alloy release as well

DaniPopes (approved):

DaniPopes (approved):

PR #870, opened 2025-02-10
feat(primitives): improve rand implementations, use `thread_rng` when available

DaniPopes:


Cargo.toml
```diff
@@ -57,7 +57,7 @@ quote = "1.0"
 syn = "2.0"
 
 cfg-if = "1.0.0"
-derive_more = { version = "1.0", default-features = false }
+derive_more = { version = "2.0", default-features = false }
```
mattsse, line 60: making a mental note here that this now exists

mattsse (approved):
makes sense

PR #874, opened 2025-02-14
feat: erc2098 signature representation

### LINKED ISSUE #873: [Feature] Add compact signature representation for `PrimitiveSignature`
### Component

primitives

Some contracts can verify using ERC-2098 signatures which saves on calldata. It would be nice to have a method to encode `PrimitiveSignature` for ERC-2098.

The representation simply stores `yParity` in the top bit of `s`, and the full encoding would be `r (256 bits) | y (1 bit) | s (255 bits)`

### Additional context

See https://eips.ethereum.org/EIPS/eip-2098 which also includes test cases

onbjerg:
## Motivation

This representation of the signature is slightly more compact and can be useful for on-chain signature verification as it saves calldata.

Closes #873 

## Solution

Add a helper to generate such a representation 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/primitive_sig.rs
+++ b/crates/primitives/src/signature/primitive_sig.rs
@@ -153,6 +153,8 @@ impl PrimitiveSignature {
     /// Normalizes the signature into "low S" form as described in
     /// [BIP 0062: Dealing with Malleability][1].
     ///
+    /// If `s` is already normalized, returns `None`.
+    ///
     /// [1]: https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki
     #[inline]
     pub fn normalize_s(&self) -> Option<Self> {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
should we add a constructor `from_erc2098` as well?

crates/primitives/src/signature/primitive_sig.rs
```diff
@@ -278,6 +278,25 @@ impl PrimitiveSignature {
...
     }
 
+    /// Returns the ERC-2098 compact representation of this signature.
+    ///
+    /// The first 32 bytes are the `r` value, and the next 32 bytes are the `s` value with `yParity`
+    /// in the top bit of the `s` value, as described in ERC-2098.
+    ///
+    /// See <https://eips.ethereum.org/EIPS/eip-2098>
+    #[inline]
+    pub fn as_erc2098(&self) -> [u8; 64] {
```
klkvr, line 288: should this also invoke `self.normalize_s` just to make sure that it always works correctly even for not non-EIP2 signatures?

PR #883, opened 2025-02-19
feat(`sol!`)!: gen unit/tuple structs for errors, calls, events with 0/1 param

yash-atreya:
## Motivation

Closes #878 
Closes #879 via #885
Closes #880 via #884

## Solution

- Generate a unit struct for error with no params and tuple struct for error with exactly one param
```rust
sol! {
   error SomeError();
   error AnotherError(uint256); // unnamed param
   error YetAnotherError(uint256 a); // named param

}
// Generates
pub struct SomeError;

// Expanded as tuple struct if single unnamed param
// The idea is to remove accessing the inner filed as `another_err._0`.
pub struct AnotherError(pub U256); 

pub struct YetAnotherError { a: U256 }; // Expanded as a struct with named fieled, same as previous bindings.

// Previous
pub struct SomeError {};

pub struct AnotherError { _0: U256 };

pub struct YetAnotherError { a: U256 } 
```
- Bindings for errors with 2 or more params are unchanged.

- Reviewer Note: #884, #885 were stacked over this and have been merged into this

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #885, opened 2025-02-21
fix(`sol!`)!: gen unit struct for no param events

### LINKED ISSUE #879: [Feature] feat(`sol!`)!: improve event bindings
### Component

sol! macro

```rust
sol! {
    contract Errors {
        event SomeEvent();
    }
}
```

Currently generated bindings:

```rust
struct SomeEvent {}
```

Proposed bindings, similar to what is proposed in #878 and #855 

```rust
// Unit struct NOT empty struct
struct SomeEvent;
```

Bindings for events with params remain unchanged.

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #879 
Stacked over #883 

## Solution

- If event does not contain any params, generate a unit struct instead of an empty struct.

```rust
sol! {
   event Empty();
}

// New 
pub struct Empty;

// Previous 
pub struct Empty {};
```

Bindings for events with params remain unchanged.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro-expander/src/expand/event.rs
+++ b/crates/sol-macro-expander/src/expand/event.rs
@@ -146,14 +146,24 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, event: &ItemEvent) -> Result<TokenStream>
 
     let alloy_sol_types = &cx.crates.sol_types;
 
+    let event_struct = if event.parameters.is_empty() {
+        quote! {
+            pub struct #name;
+        }
+    } else {
+        quote! {
+            pub struct #name {
+                #(#fields,)*
+            }
+        }
+    };
+
     let tokens = quote! {
         #(#attrs)*
         #doc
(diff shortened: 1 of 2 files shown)


crates/sol-macro-expander/src/expand/event.rs
```diff
@@ -146,14 +146,24 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, event: &ItemEvent) -> Result<TokenStream>
 
     let alloy_sol_types = &cx.crates.sol_types;
 
+    let event_struct = if event.parameters.is_empty() {
+        quote! {
+            pub struct #name;
```
mattsse, line 151: should we derive default in these cases?

mattsse (approved):

DaniPopes:
I don't think we should do this, it doesn't make that much of a difference to keep it consistent when there are no fields vs 1 or more fields.

DaniPopes (approved):

Issue #738, opened 2024-09-15
[Feature] Ability to specify a dynamic ABI file path for `sol!`


kayabaNerve:
### Component

sol! macro

### Describe the feature you would like

I built my contracts to `OUT_DIR`. `sol!(MyContract, concat!(env!("OUT_DIR"), "/contracts/MyContract.abi"))` fails to compile with "expected ident". Reviewing the macro myself, the macro appears to be looking for a string literal for the path (preventing usage of so-derived paths).

This can be worked around by hooking into `alloy-sol-macro-expander` directly from a build script. That crate has weaker semver guarantees and building custom expanders is largely undocumented (though it was a small amount of code once I worked it out). I understand if this is closed as won't fix accordingly, I just wanted to explain why I thought the issue at least worth filing.

michcio1234:
@kayabaNerve, could you please explain with more details how can one work around it? 
I have exactly the same usecase, but I have no experience in macro development, and unfortunately I don't get what you meant by "hooking into `alloy-sol-macro-expander` directly from a build script".

kayabaNerve:
https://github.com/serai-dex/serai/blob/5a7b815e2e8a6e1682a5eb5b653fb856befa24ba/processor/ethereum/router/build.rs#L5-L19

DaniPopes:
This should be straight forward to implement using [macro-string](https://github.com/dtolnay/macro-string)

Issue #659, opened 2024-06-12
[Feature] sol! attrs bytecode_path and deployed_bytecode_path


kpp:
### Component

sol! macro

### Describe the feature you would like

Now sol! macro accepts only literal hex for bytecode: `#[sol(abi, bytecode = "HEX")]`. It would be nice to be able to pass a path to .bin file with HEX bytecode inside and include it at compile time as it's done for json abi.

nadtech-hub:
✋

DaniPopes:
You will be able to specify `bytecode = include_str!(concat!(env!()))` etc. with https://github.com/alloy-rs/core/pull/898.

PR #898, opened 2025-03-10
feat(sol-macro): allow standard library macros for string literals

DaniPopes:
Use [macro-string](https://docs.rs/macro-string) to allow standard library macros such as concat!, env!, and include_str! wherever a string literal is expected in the macro input.

Closes https://github.com/alloy-rs/core/issues/738.
Closes https://github.com/alloy-rs/core/issues/659.


mattsse (approved):

yash-atreya (approved):
lgtm!

PR #901, opened 2025-03-12
fix(`sol!`): pass correct call_struct to call_builder in expansion

yash-atreya:
## Motivation

After #884, call types are expanded depending on the number of params.

This change isn't propagated to `rpc` expansion code where call_builder is called.

## Solution

- Generate the correct call struct like #884
- Pass it to `call_builder` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro-expander/src/expand/contract.rs
+++ b/crates/sol-macro-expander/src/expand/contract.rs
@@ -987,20 +987,14 @@ fn call_builder_method(f: &ItemFunction, cx: &ExpCtxt<'_>) -> TokenStream {
     let doc = format!("Creates a new call builder for the [`{name}`] function.");
 
     let call_struct = if f.parameters.is_empty() {
-        quote! {
-            pub struct #call_name;
-        }
+        quote! { #call_name }
     } else if f.parameters.len() == 1 && f.parameters[0].name.is_none() {
-        let ty = cx.expand_rust_type(&f.parameters[0].ty);
-        quote! {
-            pub struct #call_name(pub #ty);
-        }
+        let param_name = quote! { _0 };
+        quote! { #call_name(#param_name) }
     } else {
         let call_fields = param_names1.clone();
         quote! {
-            pub struct #call_name {
-                #(#call_fields),*
-            }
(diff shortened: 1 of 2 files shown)


crates/sol-macro-expander/src/expand/contract.rs
```diff
@@ -983,13 +983,30 @@ fn call_builder_method(f: &ItemFunction, cx: &ExpCtxt<'_>) -> TokenStream {
...
+            pub struct #call_name(pub #ty);
+        }
+    } else {
+        let call_fields = super::expand_fields(&f.parameters, cx);
+        quote! {
+            pub struct #call_name {
+                #(#call_fields),*
+            }
+        }
+    };
```
mattsse, line 1005: hmm, unclear why we need this here, do you have an example?

are we introducing new types here to group tuples together?

this previously didn't generate new structs and used `call_name` directly, why do we need a new struct here?

how will this be used?

mattsse (commented):
some questions

crates/sol-macro-expander/src/expand/contract.rs
yash-atreya: > this previously didn't generate new structs and used call_name directly, why do we need a new struct here?

Fixed above in [a6d20b5](https://github.com/alloy-rs/core/pull/901/commits/a6d20b58dcee00c0a8165e101637cc16468fca53)

mattsse: > struct balanceOfCall(Address); // Because there is only one param and it is unnamed.

this should be `struct balanceOfCall(pub Address);`?

yash-atreya: > > struct balanceOfCall(Address); // Because there is only one param and it is unnamed.
> 
> this should be `struct balanceOfCall(pub Address);`?

Yes, fixed above

mattsse (commented):
makes sense to me now

pending @DaniPopes

DaniPopes (approved):

PR #903, opened 2025-03-12
fix(`sol-expander`): map `self` to `this` in codegen

### LINKED ISSUE #902: [Bug] expander fails on keyword `self`
### Component

sol! macro

0.8.22

### Operating System

None

### Describe the bug

Ref https://github.com/foundry-rs/foundry/issues/10055

If keyword `self` is used, codegen would fail with `expected identifier, found keyword self` 

for. e.g 
a function with param named `self` as 

`function evaluateVanishingPoly(EvalDomain memory self, uint256 zeta)`

yash-atreya:
## Motivation

Closes #902 

## Solution

Rust does not accept the raw identifier for `self` i.e `r#self`. See: https://internals.rust-lang.org/t/raw-identifiers-dont-work-for-all-identifiers/9094/4

- Map `self` to `this`.

```rust
sol! {
  function something(address self) returns (uint256);  // Previously failed to compile
}

// Now generates
pub struct somethingCall { this: Address };
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/syn-solidity/src/ident/mod.rs
+++ b/crates/syn-solidity/src/ident/mod.rs
@@ -107,6 +107,15 @@ impl SolIdent {
 
         if matches!(s, "_" | "self" | "Self" | "super" | "crate") {
             new_raw = false;
+
+            // `self` renamed to `this` as `r#self` is not accepted by rust.
+            // See: <https://internals.rust-lang.org/t/raw-identifiers-dont-work-for-all-identifiers/9094/4>
+            if matches!(s, "self") {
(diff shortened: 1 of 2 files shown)


mattsse (approved):

DaniPopes:
Note that this also modifies the signature of a function, so `function self(uint);` would be interpreted as `this(uint256)` instead of `self(uint256)`; very unlikely, and don't think we can fix this

PR #905, opened 2025-03-13
Yash/fix 826

### LINKED ISSUE #826: [Feature] Avoid conflicting functions with common trait impls on UDT types
### Component

sol-types, sol! macro, syn-solidity

On UDT types, currently there is a impl that converts to and from the underlying.
```rust
                /// Convert from the underlying value type.
                #[inline]
                pub const fn from(value: #underlying_rust) -> Self {
                    Self(value)
                }

                /// Return the underlying value.
                #[inline]
                pub const fn into(self) -> #underlying_rust {
                    self.0
                }
```
However these functions conflict with regular `into` and `from` trait impl's. Currently this will just render them useless (the impl on the struct will always override the trait impls). I recommend changing the naming to, `from_underlying` and `into_underlying` 

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #826 

## Solution

- Rename `from` and `into` methods in user-defined type bindings to `from_underlying` to `into_underlying`.
- impl `From<underlying_type>` for `udt`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro-expander/src/expand/udt.rs
+++ b/crates/sol-macro-expander/src/expand/udt.rs
@@ -69,13 +69,13 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, udt: &ItemUdt) -> Result<TokenStream> {
 
                 /// Convert from the underlying value type.
                 #[inline]
-                pub const fn from(value: #underlying_rust) -> Self {
+                pub const fn from_underlying(value: #underlying_rust) -> Self {
                     Self(value)
                 }
 
                 /// Return the underlying value.
                 #[inline]
-                pub const fn into(self) -> #underlying_rust {
+                pub const fn into_underlying(self) -> #underlying_rust {
                     self.0
                 }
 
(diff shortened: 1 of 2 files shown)


crates/sol-types/tests/macros/sol/json.rs
```diff
@@ -54,7 +54,11 @@ fn seaport() {
     let _ = constructorCall { conduitController: Address::ZERO };
 
     // BasicOrderType is a uint8 UDVT
-    let _ = BasicOrderType::from(0u8);
+    let _o1 = BasicOrderType::from(0u8);
+
+    let _o2: BasicOrderType = 0.into();
+
+    assert!(matches!(_o1, _o2));
```
DaniPopes, line 61: this matches doesn't work the way you expect it to, it always returns true

crates/sol-macro-expander/src/expand/udt.rs
```diff
@@ -94,6 +94,13 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, udt: &ItemUdt) -> Result<TokenStream> {
                 }
             }
 
+            #[automatically_derived]
+            impl From<#underlying_rust> for #name {
```
DaniPopes, line 98: needs the other way too, From<name> for underlying

DaniPopes (changes requested):

crates/sol-types/tests/macros/sol/json.rs
yash-atreya: [417bd23](https://github.com/alloy-rs/core/pull/905/commits/417bd233d68316f8a04076cdc807cfed53975d54)

crates/sol-macro-expander/src/expand/udt.rs
yash-atreya: [3ac2bb9](https://github.com/alloy-rs/core/pull/905/commits/3ac2bb923eda1ece0da8477900e8f60844c948fe)

DaniPopes (approved):

PR #907, opened 2025-03-13
feat(primitives): remove `From<String> for Bytes`

DaniPopes:
Remove the `From<String> for Bytes` implementation to allow using `Bytes` with `clap` by default.

Reference: https://docs.rs/clap/latest/clap/macro.value_parser.html
`From<String>` takes higher precedence than `FromStr`.


yash-atreya (commented):
I think you branched off the wrong branch? https://github.com/alloy-rs/core/pull/907/commits/f323d77e010f404ca31f4dc6992e14cf06b81273 is included again in this PR

DaniPopes:
Rebased

Issue #536, opened 2024-02-21
[Feature] Improve SollCall return type encoding


rouzwelt:
### Component

sol-types, sol! macro

### Describe the feature you would like

right now when we have `somethingReturn` and want to potentially abi encode it back, there is no direct way, one way is to use `somethingCall::abi_encode_returns()` but this wont accept the somethingReturn value directly and throws errors for impl bounds are not met.
a workaround is to deconstruct the somethingReturn value when calling the abi encode like:
`somethingCall::abi_encode_returns(&(somethingReturn.field1, somethingReturn.field2, ...))`

but this is not a really good API, so an improvement would be appreciated so we can just abi encode a return type as normally we encode any SollCall value.

PR #909, opened 2025-03-17
feat(sol-macro)!: improve call return encoding

DaniPopes:
Changes `SolCall::abi_encode_returns` to accept `Self::Return` instead of the tuple.

Fixes https://github.com/alloy-rs/core/issues/536.


yash-atreya (approved):
nice, lgtm!

PR #911, opened 2025-03-17
Yash/to sol config

### LINKED ISSUE #908: feat(`json-abi`): configuration for generating structs in same namespace as the interface
### Component

json-abi

Ref: https://github.com/foundry-rs/foundry/issues/9960

After `0.8.1`, structs are written in a separate library as highlighted in the linked issue.

Add a way to configure this and write the structs in interface itself. 

Relevant code:
https://github.com/alloy-rs/core/blob/5b29eb74afb26417704833a0985a2e35bd218ee6/crates/json-abi/src/to_sol.rs#L154-L175

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #908 

## Solution

- Adds `types_in_interface: bool` key in `ToSolConfig` 
- Sets whether to generate structs or any other types in the same interface as the contract and not in a separate library. Default: `false`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-abi/src/to_sol.rs
+++ b/crates/json-abi/src/to_sol.rs
@@ -1,4 +1,5 @@
 use crate::{
+    alloc::string::ToString,
     item::{Constructor, Error, Event, Fallback, Function, Receive},
     EventParam, InternalType, JsonAbi, Param, StateMutability,
 };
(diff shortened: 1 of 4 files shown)


crates/json-abi/src/to_sol.rs
```diff
@@ -20,6 +20,9 @@ pub struct ToSolConfig {
     print_constructors: bool,
     enums_as_udvt: bool,
     for_sol_macro: bool,
+    /// Default is false. If set to `true`, any types part of some other interface/library are
+    /// generated in the same interface as the contract and not in a separate library.
+    types_in_interface: bool,
```
mattsse, line 25: these docs make sense, but I wonder if we can be a bit more explicit with the field name here, but since this is an advanced setting this seems fine

mattsse (approved):
makes sense, pending @DaniPopes

crates/json-abi/src/to_sol.rs
DaniPopes, line 25: maybe a better name like one_contract

effectively what this does is ignore the original contract.name paths and just put everything in the same contract; 
this breaks if there are structs with the same name
should be mentioned in the docs

yash-atreya: [7ebb660](https://github.com/alloy-rs/core/pull/911/commits/7ebb660946f3b60a3c08f0836a3528af2c23981a)

crates/json-abi/src/to_sol.rs
```diff
@@ -20,6 +20,11 @@ pub struct ToSolConfig {
     print_constructors: bool,
     enums_as_udvt: bool,
     for_sol_macro: bool,
+    /// Default is false. If set to `true`, any types part of some other interface/library are
+    /// generated in one contract.
+    ///
+    /// This breaks if there are structs with the same name in different interfaces.
```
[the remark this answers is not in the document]
yash-atreya: [9e2b94c](https://github.com/alloy-rs/core/pull/911/commits/9e2b94c80111a96c2b97844ef1b93a50312dd7db)

DaniPopes (approved):

PR #913, opened 2025-03-17
Make PrimitiveSignature::new a const fn

kpp:
Port https://github.com/alloy-rs/core/pull/810 from Signature to PrimitiveSignature (no idea how it was lost).


mattsse (approved):
odd

this can indeed be const

pending @DaniPopes

DaniPopes:
Sorry about that. This is now released in [0.8.24](https://github.com/alloy-rs/core/releases/tag/v0.8.24).

Issue #876, opened 2025-02-18
[Feature] Implement ToSql and FromSql traits from diesel to Signed and FixedBytes types


the-solid-box:
### Component

primitives

### Describe the feature you would like

Now that ruint has implemented diesel's FromSql and ToSql traits, we can now implement them for alloy's primitive types as well.

the-solid-box:
If that's something you'd consider merging, I can work on a PR in the next 2 to 3 weeks, unless someone else wants to take care of it.

mattsse:
I think since ruint has diesel,sql support we can do the same here, should be nice to have

yash-atreya:
@the-solid-box assigned this to you!
Sorry about the misassignment @nadtech-hub

jenpaff:
> [@the-solid-box](https://github.com/the-solid-box) assigned this to you! Sorry about the misassignment [@nadtech-hub](https://github.com/nadtech-hub)

@the-solid-box are you still working on this?

the-solid-box:
> > [@the-solid-box](https://github.com/the-solid-box) assigned this to you! Sorry about the misassignment [@nadtech-hub](https://github.com/nadtech-hub)
> 
> [@the-solid-box](https://github.com/the-solid-box) are you still working on this?

Oh I'm sorry, I missed the notification!
I also hadn't seen the assignation.

I should be able to have this done in the next week.

jenpaff:
> > > [@the-solid-box](https://github.com/the-solid-box) assigned this to you! Sorry about the misassignment [@nadtech-hub](https://github.com/nadtech-hub)
> > 
> > 
> > [@the-solid-box](https://github.com/the-solid-box) are you still working on this?
> 
> Oh I'm sorry, I missed the notification! I also hadn't seen the assignation.
> 
> I should be able to have this done in the next week.

sure, feel free to let us know if something gets in the way

PR #915, opened 2025-03-18
feat(primitives): supporting diesel @ 2.2

sinasab:
## Motivation

This PR adds diesel support for `alloy-primitives`, covering the following:
- add diesel dep, gated behind new `diesel` feature
- bump ruint dep by a patch version to include its diesel support, also gated behind the `diesel` feature
- manually implementing diesel-related traits where needed, similar to [the PR adding diesel support to ruint](https://github.com/recmo/uint/pull/404)
- macro-deriving diesel-related traits where possible, for `FixedBytes`, `Address` (via the macro that creates it), and `PrimitiveSignature`
- `PrimitiveSignature` encoded via erc2098

I had most of this code already written for another project and wanted to contribute it upstream. I noticed issue #876 had some activity but didn't see any indications of a PR so I hope this doesn’t overlap with anyone else’s efforts! 🫡 

I've posted a public example showing basic usage via diesel's postgres support [here](https://github.com/sinasab/diesel_alloy_test) in case it's useful; happy to provide any more details to get this over the line! 😁 

## Solution

The solution handles serializing/deserializing the common types exposed in `alloy-primitives` via diesel's `sql_types::Binary`. `ToSql` and `FromSql` need to be manually implemented, but some of the other traits can be derived either directly with macros `diesel` provides or via the helper lib, `diesel-derive-newtype`. 

Note that the implementation in [ruint](https://github.com/recmo/uint/blob/main/src/support/diesel.rs) inlines all of the code generated from the derive macros; I can do that here too, but figured the macros were cleaner to start with.

Closes https://github.com/alloy-rs/core/issues/876.

### DIFF
--- a/crates/primitives/src/bits/macros.rs
+++ b/crates/primitives/src/bits/macros.rs
@@ -69,7 +69,6 @@ macro_rules! wrap_fixed_bytes {
                 $extra_derives,
             )*
         )]
-        #[cfg_attr(feature = "diesel", derive(diesel_derive_newtype::DieselNewType))]
         #[repr(transparent)]
         $vis struct $name(#[into_iterator(owned, ref, ref_mut)] pub $crate::FixedBytes<$n>);
 
@@ -185,6 +184,7 @@ macro_rules! wrap_fixed_bytes {
         $crate::impl_allocative!($name);
         $crate::impl_arbitrary!($name, $n);
         $crate::impl_rand!($name);
+        $crate::impl_diesel!($name, $n);
 
         impl $name {
             /// Array of Zero bytes.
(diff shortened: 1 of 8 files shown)


mattsse (approved):
reasonble, since ruint also has this.
all of this lgtm

pending @DaniPopes

crates/primitives/src/diesel.rs
```diff
@@ -0,0 +1,57 @@
...
+where
+    for<'c> Db: Backend<BindCollector<'c> = RawBytesBindCollector<Db>>,
+{
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Db>) -> SerResult {
+        out.write_all(&self[..])?;
+        Ok(IsNull::No)
+    }
+}
+
+impl<const BITS: usize, Db: Backend> FromSql<Binary, Db> for FixedBytes<BITS>
```
DaniPopes, line 27: these are BYTES not BITS

crates/primitives/src/diesel.rs
```diff
@@ -0,0 +1,57 @@
...
+where
+    *const [u8]: FromSql<Binary, Db>,
+{
+    fn from_sql(bytes: Db::RawValue<'_>) -> DeserResult<Self> {
+        let bytes: *const [u8] = FromSql::<Binary, Db>::from_sql(bytes)?;
+        let bytes = unsafe { &*bytes };
+        Ok(Self::from_slice(bytes))
```
DaniPopes, line 34: This panics if the length is incorrect, use `try_from`

crates/primitives/src/diesel.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl<Db: Backend> FromSql<Binary, Db> for PrimitiveSignature
+where
+    *const [u8]: FromSql<Binary, Db>,
+{
+    fn from_sql(bytes: Db::RawValue<'_>) -> DeserResult<Self> {
+        let bytes: *const [u8] = FromSql::<Binary, Db>::from_sql(bytes)?;
+        let bytes = unsafe { &*bytes };
+        Ok(Self::from_erc2098(bytes))
```
DaniPopes, line 55: same, need to check the length here `bytes.len() == 64`

crates/primitives/src/bits/macros.rs
```diff
@@ -65,6 +65,7 @@ macro_rules! wrap_fixed_bytes {
                 $extra_derives,
             )*
         )]
+        #[cfg_attr(feature = "diesel", derive(diesel_derive_newtype::DieselNewType))]
```
DaniPopes, line 68: this won't work if used outside this crate, please see how other traits are implemented below with a cfg(feature) macro_rules

DaniPopes (changes requested):

crates/primitives/src/diesel.rs
sinasab: Whoops, thanks for the catch; fixed! 🙏

crates/primitives/src/diesel.rs
sinasab: Fixed! 🫡

sinasab:
I appreciate the reviews 🙏 

@DaniPopes I went ahead and pushed up a commit covering each of your comments. Specifically:
- minor updates in `diesel.rs` as per the comments
- removing `diesel_derive_newtype` dep and inlining implementations for fixed_byte macro via the macro_rules pattern used throughout that file

Please take another look when you get the chance! 😁 

I also sanity-checked that everything still works in my demo repo; please let me know if there's any more feedback I can address 🫡

crates/primitives/src/bits/macros.rs
```diff
@@ -669,6 +670,117 @@ macro_rules! impl_arbitrary {
     ($t:ty, $n:literal) => {};
 }
 
+#[doc(hidden)]
+#[macro_export]
+#[cfg(feature = "diesel")]
+macro_rules! impl_diesel {
+    ($t:ty, $n:literal) => {
+        use $crate::private::diesel::{
```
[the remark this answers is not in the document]
sinasab: Done! 🫡

sinasab:
Whoops, had a clippy fix fall through the cracks from the `PrimitiveSignature` deprecation. Should be fixed now!

DaniPopes (approved):

PR #918, opened 2025-03-21
fix: do not rely on bytes dependency in `wrap_fixed_bytes!`

klkvr:
## Motivation

`wrap_fixed_bytes` can't be used without `bytes` as a dependency.

there's similar issue with `derive_more` but I'm unsure how we can handle this given that it seems to rely on `derive_more` being available internally

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/bits/macros.rs
+++ b/crates/primitives/src/bits/macros.rs
@@ -4,8 +4,9 @@
 /// This functionally creates a new named `FixedBytes` that cannot be
 /// type-confused for another named `FixedBytes`.
 ///
-/// **NOTE:** This macro currently requires `#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]`
-/// at the top level of the crate where it is used.
+/// **NOTE:** This macro currently requires:
+/// - `#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]` at the top level of the crate.
+/// - The `derive_more` crate in scope.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):
wdym about derive_more? it's also available in $crate::private

can you also add a wrap_fixed_bytes! call to alloy-core-sol-test? this shouldve caught it

klkvr:
> wdym about derive_more? it's also available in $crate::private

if you write
```rust
#[derive(alloy::primitives::private::derive_more::AsRef)]
struct MyStruct(u8);
```

it will expand to 
```rust
impl derive_more::core::convert::AsRef<u8> for MyStruct {
    #[inline]
    fn as_ref(&self) -> &u8 {
        &self.0
    }
}
```

causing issues if `derive_more` is not added as a dep, unsure if there's a way to work around this

klkvr:
ah I guess it can be solved by adding `use $crate::private::derive_more`

klkvr:
there are also some `doc(cfg)` issues, unsure how to handle. do we need them? looks like they imply that crates using the macro would have those features which might not be true?

DaniPopes:
makes sense, i think the "use derive_more" is ok unless it causes other errors

DaniPopes:
ah doc cfg requires the feature at the top level, it's unfortunate because otherwise the feature does not show up in rustdoc because of how the methods are implemented using macros

DaniPopes:
It looks like `#[feature(doc_cfg, doc_auto_cfg)]` are on path to stabilization, and in general are a good features to enable for docs, so I wouldn't mind this as a requirement to users of this macro.

PR #919, opened 2025-03-24
fix(primitives): Remove undefined behavior in FixedBytes

dist1ll:
## Motivation

Currently, the code for generating random `FixedBytes` primitives is incorrect. It calls `assume_init` directly on an uninitialized array, which invokes immediate undefined behavior.

Note that it doesn't matter that we technically never explicitly read from the array before randomizing. An uninitialized byte array is always invalid and violates Rust rules. Refer to the [docs for MaybeUninit](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html).

## Solution

We should follow the documentation of [`MaybeUninit::assume_init`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.assume_init) and make sure every element was written to. Since we're initializing arrays, I used `[MaybeUninit<u8>; N]` (idiomatic for Rust), together with `array_assume_init` (currently unstable, I added a cfg_attr-gated #![feature]).

Luckily `getrandom` provides a function for randomizing uninitialized arrays, so I could use the MaybeUninit array directly. For `rand`, I couldn't find an equivalent, so I implemented the `Fill` trait manually via a wrapper.

## PR Checklist

- [ ] Added Tests (covered by existing tests)
- [x] Added Documentation (no pub doc changes, only added safety comments)
- [ ] Breaking changes (none)

### DIFF
--- a/crates/primitives/src/bits/fixed.rs
+++ b/crates/primitives/src/bits/fixed.rs
@@ -372,7 +372,9 @@ impl<const N: usize> FixedBytes<N> {
     #[inline]
     #[track_caller]
     pub fn random() -> Self {
-        Self::try_random().unwrap_or_else(|e| panic!("failed to fill with random bytes: {e}"))
+        let mut bytes = Self::ZERO;
+        bytes.randomize();
+        bytes
     }
 
     /// Tries to create a new [`FixedBytes`] with the default cryptographic random number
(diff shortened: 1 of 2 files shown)


DaniPopes (changes requested):
Makes sense. We should just use `Self::ZERO` to make this change easier and not require nightly features, even if it zero's the memory unnecessarily.

Can you also make a PR to https://github.com/recmo/uint/ with the same change (src/support/rand*.rs)?

DaniPopes (approved):

DaniPopes:
Going to go ahead and make the changes myself because we're looking to publish a new version soon, thanks for the contribution.

PR #920, opened 2025-03-24
chore: add hash_ref function to sealed.rs

shane-moore:
## Changes

- Introduced a new `hash_ref()` method on `Sealed<T>` to expose a reference to the internal `seal` hash.
- This addition enables consumers to access the hash without incurring a copy, which is useful when only a reference is needed.

## Motivation

This change supports [reth#15250](https://github.com/paradigmxyz/reth/pull/15250), where `OpTxEnvelope::Deposit` will implement `SignedTransaction`. Since the `hash()` method on `Sealed<TxDeposit>` returns by value, `hash_ref()` allows returning a borrowed hash to satisfy trait requirements (`&TxHash`)

### DIFF
--- a/crates/primitives/src/sealed.rs
+++ b/crates/primitives/src/sealed.rs
@@ -107,6 +107,12 @@ impl<T> Sealed<T> {
         self.seal
     }
 
+    /// Get the hash.
+    #[inline(always)]
+    pub const fn hash_ref(&self) -> &B256 {
+        &self.seal
+    }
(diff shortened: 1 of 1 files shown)


shane-moore:
@mattsse, could you take a look when get a chance? 🙏 

This addition is in support of https://github.com/paradigmxyz/reth/issues/15221

DaniPopes (approved):

PR #863, opened 2025-01-31
Yash/rm validate bool

yash-atreya:
## Motivation

Currently, in `abi_decode_*` methods of sol type traits (`SolEnum`, `SolType`, `SolValue`, `SolCall`, `SolInterface`, `SolEvents`), we expose a `validate: bool` arg to bypass type-checking while decoding. We should always resort to type-checking. 

This type-checking ensures that bytes conform to expected type limitations and that the decoded values can be re-encoded to an identical byte-string.

## Solution

- Remove the `validate: bool` from `abi_decode_*` fns of various sol type traits. 
- Address the consequent breaking changes in the `sol!` expansions.
- Fix tests.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

PR #928, opened 2025-04-16
feat: add validated variants to ABI decoding methods

### LINKED ISSUE #927: [Feature] Add abi_decode*_validate methods
### Component

sol-types

The `validate` boolean parameter was removed in https://github.com/alloy-rs/core/pull/863, but we still want to expose the validation logic for users who need it. This is still present with `SolType::type_check`.

Add `abi_decode*_validate` methods to relevant traits and types to expose this functionality again.

### Additional context

_No response_

harsh-ps-2003:
## Motivation
The fundamental motivation behind these changes is to re-introduce ABI decoding validation that was effectively removed by default in PR #863 

This PR removed the validate: bool argument from decoding functions (abi_decode*). The reason was that the existing validation logic was sometimes too strict and incorrectly rejected valid ABI-encoded data found in real-world scenarios. Removing the flag made the default decoding behavior non-validating (or less strict).

However, performing strict ABI validation is still a desirable feature for users who want to ensure data adheres precisely to the expected format, potentially catching encoding errors or unexpected data structures early.

Fixes #927 

## Solution

So, instead of re-introducing the boolean flag (which complicates the function signatures), the chosen approach was to add separate, parallel methods specifically for validated decoding. These new methods have _validate appended to their names.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/sol-types/src/types/error.rs
```diff
@@ -286,6 +314,11 @@ impl SolError for Panic {
     fn abi_encoded_size(&self) -> usize {
         32
     }
+
+    #[inline]
+    fn abi_decode_raw_validate(data: &[u8]) -> Result<Self> {
```
DaniPopes, line 319: Why are these overridden here?

crates/sol-macro/src/lib.rs
DaniPopes, line 1: File needs reverting to satisfy formatter; I have encountered this in the past, I don't really know why these changes happen

DaniPopes (changes requested):

harsh-ps-2003:
Can you retake a look @DaniPopes ?

DaniPopes (approved):
Thanks!

PR #933, opened 2025-04-19
docs: fixed broken link

cypherpepe:
## Motivation

The relative link to the `abi` module in `README.md` was broken — GitHub couldn't resolve `abi` on its own.  
Switched it to `src/abi` so it correctly points to the actual folder.

## Solution

...

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (commented):
we include this in lib.rs like:

https://github.com/alloy-rs/core/blob/74717c0beb453adf093affd2c3ffc93c89ad9327/crates/sol-types/src/lib.rs#L1

so from that pov the link is accurate

PR #943, opened 2025-04-29
feat(primitives): add BitX<&Self> for FixedBytes

### LINKED ISSUE #941: [Feature] Add BitX operations for &Self
### Component

primitives

all of these impls operate on &rhs

https://github.com/alloy-rs/core/blob/8c7229d97317bbd4e8a39e829e76b1a716c3caad/crates/primitives/src/bits/fixed.rs#L264-L270

we can add support for

```
 impl<const N: usize> ops::BitAndAssign<&Self> for FixedBytes<N> { 
```

for all Bit ops

## TODO
* add impls for &Self
* delegate current uses like `self.bitand_assign(&rhs)`

### Additional context

_No response_

Rimeeeeee:
fixes #941


crates/primitives/src/bits/fixed.rs
```diff
@@ -251,21 +255,44 @@ impl<const N: usize> fmt::UpperHex for FixedBytes<N> {
...
+    }
+}
+
+impl<const N: usize> ops::BitXorAssign<&Self> for FixedBytes<N> {
+    #[inline]
+    fn bitxor_assign(&mut self, rhs: &Self) {
+        // Note: `slice::Iter` has better codegen than `array::IntoIter`
+        iter::zip(self, rhs).for_each(|(a, b)| *a ^= *b);
+    }
+}
```
mattsse, line 280: these look good, can we group them to the existing `BitX` implements so that we have all of them together.

the old impls of those traits can now just call `self.bitxor_assign(&rhs)` this way we can reuse some code

mattsse (changes requested):

crates/primitives/src/bits/fixed.rs
```diff
@@ -251,39 +255,47 @@ impl<const N: usize> fmt::UpperHex for FixedBytes<N> {
...
-    fn bitand_assign(&mut self, rhs: Self) {
+    fn bitxor_assign(&mut self, rhs: &Self) {
         // Note: `slice::Iter` has better codegen than `array::IntoIter`
-        iter::zip(self, &rhs).for_each(|(a, b)| *a &= *b);
+        iter::zip(self, rhs).for_each(|(a, b)| *a ^= *b);
     }
 }
 
-impl<const N: usize> ops::BitOr for FixedBytes<N> {
+impl<const N: usize> ops::BitAnd for FixedBytes<N> {
```
DaniPopes, line 282: please also add Bit*<&Self>

crates/primitives/src/bits/fixed.rs
```diff
@@ -292,16 +304,29 @@ impl<const N: usize> ops::BitXor for FixedBytes<N> {
 
     #[inline]
     fn bitxor(mut self, rhs: Self) -> Self::Output {
-        self ^= rhs;
+        self.bitxor_assign(&rhs);
```
DaniPopes, line 307: this doesn't have to be changed

mattsse (changes requested):

mattsse (approved):

PR #950, opened 2025-05-09
chore: enhance eip712 string parser to canonicalize inputs

0xrusowsky:
This PR adds a `fn canonicalize()` method to `EncodeType` to give more flexibility to users when defining the string representations of their types. Thanks to this new method they can be certain that they will get the canonical type definition to compute the correct hash.

before users had to pass an already canonical input:

```rs
const CANONICAL: &str =
"Transaction(Person from,Person to,Asset tx)Asset(address token,uint256 amount)Person(address wallet,string name)";
```

now they can simply pass:
```rs
const MESSY: &str = r#"
    Person(address wallet, string name) Asset(address token, uint256 amount)
    Transaction(Person from, Person to, Asset tx)
    "#;
```

it also adds a couple of check to ensure that all types used in properties have their components defined, and to ensure that there is a (single) primary type


crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -2,6 +2,8 @@
 
 // TODO: move to `sol-type-parser`
 
+use std::collections::HashSet;
```
mattsse, line 5: I think we can't use this because this breaks no-std support

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +141,91 @@ impl<'a> EncodeType<'a> {
...
+            return Err(Error::MissingType("Primary Type".into()));
+        }
+
+        let primary_idx = self.get_primary_idx()?;
+
+        // EIP-712 requires alphabetical order of the secondary types
+        let mut types = self.types.clone();
+        let mut sorted = vec![types.remove(primary_idx)];
+        types.sort_by(|a, b| a.type_name.cmp(b.type_name));
+        sorted.extend(types);
```
mattsse, line 159: this would be equivalent to `types.insert(0, types.remove(primary_idx)// do before]`

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +141,91 @@ impl<'a> EncodeType<'a> {
...
+                let type_str = prop_def.ty.span.trim();
+                let type_str = type_str.split('[').next().unwrap_or(type_str).trim();
+
+                // A type is considered a reference to another type if its name starts with an
+                // uppercase letter, otherwise it is assumed to be a basic type
+                if !type_str.is_empty()
+                    && type_str.chars().next().is_some_and(|c| c.is_ascii_uppercase())
+                {
+                    types_in_props.insert(type_str);
+                }
```
mattsse, line 188: I think the order can be swapped here because we only need to split at [ if this is uppercase?

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +141,91 @@ impl<'a> EncodeType<'a> {
...
+        let mut is_found = false;
+        for (n, ty) in self.types.iter().enumerate() {
+            if !types_in_props.contains(ty.type_name) {
+                if is_found {
+                    return Err(Error::MissingType("no primary component".into()));
+                }
+                primary = n;
+                is_found = true;
+            }
+        }
```
mattsse, line 210: hmm, isn't this more or less equivalent to something like:

iter over `self.types`
filter by `self.types.flat_map(types.prop.ty).any(prop_ty.starts_with(ty))`

mattsse (changes requested):
nits and one q

crates/dyn-abi/src/eip712/parser.rs
0xrusowsky: ohhh i forgot... i'll use `alloc::vec` instead + manually avoid duplicates

crates/dyn-abi/src/eip712/parser.rs
0xrusowsky: reduced the extra allocation with ur suggestion + made the fn take a mut ref to simplify even more

crates/dyn-abi/src/eip712/parser.rs
0xrusowsky: replaced it with something closer to ur suggestion, but still using enumerate to return the idx and easily sort later on

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +139,89 @@ impl<'a> EncodeType<'a> {
...
+    fn get_primary_idx(&self) -> Result<usize, Error> {
+        // Track all types used in component properties.
+        let mut types_in_props: Vec<&str> = Vec::new();
+
+        for ty in self.types.iter() {
+            for prop_def in ty.props.iter() {
+                let type_str = prop_def.ty.span.trim();
+                // A type is considered a reference to another type if its name starts with an
+                // uppercase letter, otherwise it is assumed to be a basic type
+                if !type_str.is_empty() && type_str.chars().next().unwrap().is_ascii_uppercase() {
```
DaniPopes, line 175:
```suggestion
                if type_str.starts_with(char::is_ascii_uppercase) {
```

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +139,89 @@ impl<'a> EncodeType<'a> {
...
+    /// The primary type is the component type that is not used as a property in any component type
+    /// definition within this set.
+    fn get_primary_idx(&self) -> Result<usize, Error> {
+        // Track all types used in component properties.
+        let mut types_in_props: Vec<&str> = Vec::new();
+
+        for ty in self.types.iter() {
+            for prop_def in ty.props.iter() {
+                let type_str = prop_def.ty.span.trim();
+                // A type is considered a reference to another type if its name starts with an
```
DaniPopes, line 173: this is a heuristic, we can also check the other fields in prop_def to see if it's a custom type?

DaniPopes (commented):
could you also check this @prestwich

0xrusowsky:
> could you also check this @prestwich 

for ctx the motivation behind the PR is the impl of a new foundry cheatcode https://github.com/foundry-rs/foundry/issues/4818

crates/dyn-abi/src/eip712/parser.rs
0xrusowsky: thanks for the suggestion, i looked into the type defs and impls more carefully and made it deterministic rather than using the heuristic

mattsse (changes requested):

prestwich:
concept ACK but you should use the resolver instead of implementing this on EncodeType

prestwich:
https://github.com/alloy-rs/core/blob/b20e2326796827cbc5ca9ff7bd037fab9ba37e93/crates/dyn-abi/src/eip712/resolver.rs#L422

0xrusowsky:
> concept ACK but this code is duplicative and you should use the resolver instead of rewriting this on EncodeType. EncodeType is a parsing aid, not intended for use other than in parsing. generally your code should not import it

i see, makes sense, thanks for the explanation.
i didn't use the resolver as i thought it was only meant for json inputs.

anyways, i think i will end up leveraging solar to implement the cheatcode. Then, users only need to pass the contract + the struct name, rather than having to manually build the string representation of the types.

we can the PR if it doesn't align with the architecture that @prestwich had in mind

prestwich:
> we can close the PR if it doesn't align with the architecture that @prestwich had in mind

to be clear, I think that having the `canonicalize` method on `EncodeType` is a good idea. I'm pointing out that you can replace the logic in the body of the function with use of the resolver, and avoid adding maintenance surface area :)

> anyways, i think i will end up leveraging solar to implement the cheatcode. Then, users only need to pass the contract + the struct name, rather than having to manually build the string representation of the types.

what's the desired experience here? what sort of code would you like to write to achieve the goal?

0xrusowsky:
oh, i see what you mean, will update to incorporate the suggestion then.

> what's the desired experience here? what sort of code would you like to write to achieve the goal?

the desired experience is for them to only need to pass the contract and struct names, which requires introspection. The idea is to use something like this:
```
vm.hashType("MyContract.MyStruct") => bytes32 typeHash
```

however, supporting the string representation of the type could also be useful + adding functionality to the resolver seems like a nice-to-have
```
vm.hashType("Person(address wallet, string name) Asset(address token, uint256 amount) Transaction(Person from, Person to, Asset tx)") => bytes32 typeHash
```

prestwich:
> oh, i see what you mean, will update to incorporate the suggestion then.
> 
> > what's the desired experience here? what sort of code would you like to write to achieve the goal?
> 
> the desired experience is for them to only need to pass the contract and struct names, which requires introspection. The idea is to use something like this:
> 
> ```
> vm.hashType("MyContract.MyStruct") => bytes32 typeHash
> ```
> 
> however, supporting the string representation of the type could also be useful + adding functionality to the resolver seems like a nice-to-have
> 
> ```
> vm.hashType("Person(address wallet, string name) Asset(address token, uint256 amount) Transaction(Person from, Person to, Asset tx)") => bytes32 typeHash
> ```

seems like a cool feature. you could actually have the cheat code context hold a resolver, and populate all types into it as they're accessed? that'd have issues if multiple types have the same name tho 🤔

0xrusowsky:
@prestwich how do u feel about this new impl?

crates/dyn-abi/src/eip712/resolver.rs
```diff
@@ -496,6 +496,24 @@ impl Resolver {
     pub fn contains_type_name(&self, name: &str) -> bool {
         self.nodes.contains_key(name)
     }
+
+    /// Returns the names of those types which do not depend on any other nodes in the resolver
+    /// graph.
+    pub fn non_dependent_types(&self) -> Vec<&str> {
```
mattsse, line 502: should we return the TypeDef here as well, so instead of keys() we do iter(), (&str, &TypeDef)

mattsse (commented):
smol q

otherwise lgtm

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +141,59 @@ impl<'a> EncodeType<'a> {
...
+        let primary = resolver.non_dependent_types();
+        if primary.len() != 1 {
+            let msg = if primary.is_empty() {
+                "primary component".into()
+            } else {
+                format!(
+                    "primary component: {}",
+                    primary.iter().map(|t| t.type_name()).collect::<Vec<_>>().join(", ")
+                )
+            };
```
mattsse, line 166: you could use a mut string then push the type name if any, but imo this is also okay

mattsse (approved):
lgtm

crates/dyn-abi/src/eip712/resolver.rs
```diff
@@ -496,6 +496,30 @@ impl Resolver {
...
+            .filter_map(
+                |(node, def)| {
+                    if !dependent_types.contains(node.as_str()) {
+                        Some(def)
+                    } else {
+                        None
+                    }
+                },
+            )
+            .collect()
```
prestwich, line 521: api nit: this could return  `impl Iterator`

prestwich (approved):
looks good thank you :)

crates/dyn-abi/src/eip712/resolver.rs
0xrusowsky: afk today, but i can update it tomorrow if u prefer it this way :)

prestwich: thanks! the reason to prefer it is that it allows the caller to decide whether to allocate or whether to iterate. by returning the iterator, we let them decide whether they need to reserve heap space or not

0xrusowsky: yes, that's better i agree! it's already implemented

DaniPopes:
Pending CI

crates/dyn-abi/src/eip712/parser.rs
```diff
@@ -139,13 +141,61 @@ impl<'a> EncodeType<'a> {
...
+        for component_type in &self.types {
+            resolver.ingest(component_type.to_owned());
+        }
+
+        // Resolve and validate non-dependent types
+        let mut non_dependent = resolver.non_dependent_types();
+
+        let first = non_dependent
+            .next()
+            .ok_or_else(|| Error::MissingType(String::from("primary component")))?;
```
mattsse, line 161:
```suggestion
            .ok_or_else(|| Error::MissingType("primary component".to_string()))?;
```

might night `alloc::string::ToString` import

PR #953, opened 2025-05-20
feat(primitives): serialize `Signed` with a compact binary representation

stalkopat:
## Motivation
Ints are currently always serialized as strings, even for binary formats. This causes quite a bit of overhead depending on the downstream usage. (For example persisting to disk)
I256::MAX with the existing serialization format uses ~85 bytes with bincode, v.s. 40 bytes with this change

## Solution
Use is_human_readable to serialize Ints as compact binary blobs instead.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

## Considerations
This change isn't backwards compatible with already serialized data that has been f.x. persisted to disk with a non human readable format. Keeping the deserialize_any in the deserialize *should* technically work, but apparently some of the self describing formats just treat strings and bytes interchangeably, which means those will face compatibility issues anyway.

Making this change has the additional benefit of now allowing f.x. bincode to work with ints (because it gets rid of the deserialize_any in the non readable path).

Json and related readable formats (which I assume is the majority of downstream usage) shouldn't be impacted.


stalkopat:
On a second thought, up until ~2^103 the string representation is actually smaller than the fixed 32 byte representation, this might do more harm than good for the vast majority of use cases. Definitely think deserialize_any should go for non human readable formats though. Just serializing as string feels awkward aswell though

PR #960, opened 2025-06-04
chore: bump to edition 2024

DaniPopes:
Based on https://github.com/alloy-rs/core/pull/959. See individual commits.


mattsse:
ah few missing things, unsafe external block and fmt

PR #963, opened 2025-06-12
fix: colon 712 identifiers

mattsse:
ref https://github.com/foundry-rs/foundry/issues/10765

: in names causing `TypeSpecifier` identifier parsing to fail, not sure how this should be solved.

there are two aspects to this issue, is a `A:B` entry valid, and how should this be resolved and hashed

@DaniPopes

### DIFF
--- a/crates/dyn-abi/src/eip712/resolver.rs
+++ b/crates/dyn-abi/src/eip712/resolver.rs
@@ -366,7 +366,7 @@ impl Resolver {
         if self.detect_cycle(type_name, &mut context) {
             return Err(Error::circular_dependency(type_name));
         }
-        let root_type = type_name.try_into()?;
+        let root_type = type_name.rsplit(':').next().unwrap().try_into()?;
         let mut resolution = vec![];
         self.linearize_into(&mut resolution, root_type)?;
         Ok(resolution)
@@ -518,6 +518,7 @@ mod tests {
     use super::*;
     use alloc::boxed::Box;
     use alloy_sol_types::sol;
+    use serde_json::json;
 
     #[test]
     fn it_detects_cycles() {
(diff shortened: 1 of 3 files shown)


prestwich:
by spec, EIP-712 type names must be valid identifiers. The solidity grammar does not permit `:` characters. This means they are illegal in EIP-712, and allowing them is not spec-compliant. See the first **definition** in this [heading](https://eips.ethereum.org/EIPS/eip-712#definition-of-typed-structured-data-%F0%9D%95%8A), as well as the [rule for identifiers in the solidity grammar](https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityLexer.Identifier)

DaniPopes:
Is skipping everything up to and including the colon the expected behavior here? How do other ethereum libraries handle this case? I'm assuming they would include everything in the identifier, not skip

crates/sol-type-parser/src/ident.rs
```diff
@@ -76,6 +76,34 @@ where
     Ok(input.next_slice(len))
 }
 
+/// Returns `true` if the given character is valid in an EIP-712 identifier.
+///
+/// Extends the standard identifier character set to include `:` for namespace support in type
+/// names.
+#[cfg(feature = "eip712")]
+#[inline]
+const fn is_eip712_id_continue(c: char) -> bool {
```
DaniPopes, line 85: how do other libraries handle parsing this? i would assume they just accept any character in which case we're just gonna have the same issue again when someone decides to add another character in their type name

DaniPopes (changes requested):

crates/sol-type-parser/src/ident.rs
0xrusowsky: both `⁠ethers.js` and ⁠`wevm/abitype` are quite permissive with type names, and allow all characters. Their main constraint is avoiding chars that would break the encoding parsers, like brackets ⁠`[]` and parentheses ⁠`()`. Based on that, we would have flexibility to expand the allowed character set.

however, i'd personally keep a stricter implementation that remains as close to the official spec as possible. Since this patch introduces custom EIP-712 parsing, it would be easy to relax these rules in the future if needed.

with that being said, obviously i defer the decision to u guys and will impl what u think it's best

cc: @mattsse @prestwich

mattsse: @DaniPopes should we remove this function or keep for now?

PR #968, opened 2025-06-23
Fix typos in documentation and test cases

VolodymyrBg:
- Corrected a typo in documentation: "divisble" → "divisible" in `crates/primitives/src/signed/int.rs`.
- Fixed typos in test inputs and expected values in `crates/sol-type-parser/src/parameter.rs`: corrected "storag" to "storage".


DaniPopes:
This is intentional

PR #970, opened 2025-06-26
feat: add native sqlx support for Address (MySQL/Postgres/SQLite) with feature gating and tests

Rollp0x:
Motivation
Add native sqlx support for alloy_primitives::Address, enabling seamless integration with MySQL, PostgreSQL, and SQLite. This allows users to use Address directly as a database field in sqlx-powered Rust projects.
Currently, there is no out-of-the-box support for this in alloy, which limits database integration scenarios.

Solution

- Implements sqlx_core::Type, Encode, and Decode for Address (feature-gated, supports MySQL/Postgres/SQLite).

- Uses sqlx_core directly to minimize dependency bloat and align with workspace style.

- Follows the style and feature gating of existing diesel/postgres support.

- Minimum supported sqlx version: 0.8.

- Includes roundtrip tests.

All integration scenarios are covered in a public test repository:
👉 [Rollp0x/alloy-primitives-sqlx-demo](https://github.com/Rollp0x/alloy-primitives-sqlx-demo)

PR Checklist
<input checked="" disabled="" type="checkbox"> Added Tests (see [integration test repo](https://github.com/Rollp0x/alloy-primitives-sqlx-demo))
<input disabled="" type="checkbox"> Breaking changes (None)

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -157,7 +157,6 @@ serde = [
     "rand?/serde",
 ]
 
-
 allocative = ["dep:allocative"]
 arbitrary = [
     "std",
--- a/crates/primitives/src/bits/macros.rs
+++ b/crates/primitives/src/bits/macros.rs
@@ -11,8 +11,6 @@
 /// # Examples
 ///
 /// ```
-/// #![cfg_attr(not(feature = "std"), no_std)]
-/// extern crate alloc;
 /// use alloy_primitives::wrap_fixed_bytes;
 ///
 /// // These hashes are the same length, and have the same functionality, but
@@ -866,13 +864,15 @@ macro_rules! impl_diesel {
 macro_rules! impl_sqlx {
     ($t:ty, $n:literal) => {
         const _: () = {
-            use alloc::vec::Vec;
-            use $crate::private::sqlx_core::{
-                database::Database,
-                decode::Decode,
-                encode::{Encode, IsNull},
-                error::BoxDynError,
-                types::Type,
+            use $crate::private::{
+                Vec,
+                sqlx_core::{
+                    database::Database,
+                    decode::Decode,
+                    encode::{Encode, IsNull},
+                    error::BoxDynError,
+                    types::Type,
+                },
             };
 
             impl<DB> Type<DB> for $t
--- a/tests/core-sol/src/lib.rs
+++ b/tests/core-sol/src/lib.rs
@@ -7,7 +7,6 @@
 #![cfg_attr(docsrs, allow(unexpected_cfgs))]
 
 use alloy_core::{primitives::wrap_fixed_bytes, sol};
-extern crate alloc;
 
 wrap_fixed_bytes! {
     struct MyFixedBytes<32>;


crates/primitives/Cargo.toml
```diff
@@ -153,6 +156,7 @@ serde = [
     "indexmap?/serde",
     "rand?/serde",
 ]
+sqlx = ["std", "sqlx/mysql", "sqlx/sqlite", "sqlx/postgres"]
```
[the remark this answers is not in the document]
Rollp0x: Sure, I’ll try to align the implementation with the ruint version and update the PR accordingly. Thanks for the suggestion!

crates/primitives/src/sqlx.rs
```diff
@@ -0,0 +1,134 @@
+//! Support for the [`sqlx`](https://crates.io/crates/sqlx) crate.
+//!
+//! Implements SQLx traits for [`Address`], allowing seamless integration with MySQL, PostgreSQL,
+//! and SQLite.
+
+use crate::Address;
+use core::str::FromStr;
+
+#[cfg(feature = "sqlx")]
+mod sqlx_impl {
```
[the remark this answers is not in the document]
Rollp0x: Aligned with ruint style, all local checks (build, test, clippy, fmt, deny) passed. Ready for review!

crates/primitives/Cargo.toml
```diff
@@ -153,6 +156,7 @@ serde = [
     "indexmap?/serde",
     "rand?/serde",
 ]
+sqlx = ["dep:sqlx-core"]
```
mattsse, line 159: this should also forward to ruint, see postgres feature

mattsse (changes requested):
nit, other than that this lgtm

comparing with https://github.com/recmo/uint/blob/main/src/support/sqlx.rs

wdyt @DaniPopes

DaniPopes:
why only address? this can be implemented for fixedbytes<n> and then in the wrap_fixed_bytes macro

mattsse:
right, @Rollp0x we can impl this for all fixedbytes

crates/primitives/Cargo.toml
Rollp0x: Got it, something like `postgres = ["std", "dep:postgres-types", "ruint/postgres"]` and `diesel = ["std", "dep:diesel", "ruint/diesel"]`

tests/core-sol/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 #![cfg_attr(docsrs, allow(unexpected_cfgs))]
 
 use alloy_core::{primitives::wrap_fixed_bytes, sol};
+extern crate alloc;
```
mattsse, line 10: do we even need this if sqlx feature also activates std feature?

crates/primitives/src/bits/macros.rs
```diff
@@ -857,6 +860,68 @@ macro_rules! impl_diesel {
     ($t:ty, $n:literal) => {};
 }
 
+#[doc(hidden)]
+#[macro_export]
+#[cfg(feature = "sqlx")]
+macro_rules! impl_sqlx {
+    ($t:ty, $n:literal) => {
+        const _: () = {
+            use alloc::vec::Vec;
```
mattsse, line 869: sqlx feature also enables std, then we should be able to use std::vec::Vec and dont need the extern crate alloc?

mattsse (commented):
only 1 q re extern crate alloc

crates/primitives/src/bits/macros.rs
Rollp0x: when we use `std::vec::Vec`, then in `tests/core-sol/src/lib.rs`,  because the `#![no_std]` restriction is used, the std environment does not exist. `cargo check --all-features` will be failed because it will enable the` sqlx` feature, which requires a std environment. I don't want to change the `#![no_std]` restriction because it might affect other tests.

Rollp0x: You are right. I use `$crate::private::Vec` instead of `alloc::vec::Vec` in `impl_sqlx!` macro, so all problems are solved. Even `tests/core-sol/src/lib.rs` does not need to add `extern crate alloc;`. The integration test link with the database is [https://github.com/Rollp0x/alloy-primitives-sqlx-demo](https://github.com/Rollp0x/alloy-primitives-sqlx-demo).

PR #977, opened 2025-07-07
feat(primitives): extend implementation of diesel's ToSql to Sqlite for FixedBytes and Address

xavierm02:
## Motivation

The implementation of `diesel::serialize::ToSql` had a trait boung `Db: Backend<BindCollector<'c> = RawBytesBindCollector<Db>>`. This works fine for the `Mysql` and `Pg` backends, but not for the `Sqlite` backend because [it has another `BindCollector`](https://docs.diesel.rs/master/diesel/backend/trait.Backend.html#impl-Backend-for-Sqlite).

## Solution

This PR replaces the trait bound by `[u8]: ToSql<Binary, Db>`, [which is implied by the previous bound](https://docs.diesel.rs/master/diesel/serialize/trait.ToSql.html#impl-ToSql%3CBinary,+DB%3E-for-%5Bu8%5D), but [also holds for `Db = Sqlite`](https://docs.diesel.rs/master/diesel/serialize/trait.ToSql.html#impl-ToSql%3CBinary,+Sqlite%3E-for-%5Bu8%5D). [The implementation is basically the same as the one given by diesel for `Vec<u8>`](https://docs.diesel.rs/master/src/diesel/type_impls/primitives.rs.html#178-186) (and other types whose content can be borrowed as a `&[u8]`).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/primitives/src/bits/macros.rs
```diff
@@ -785,7 +785,8 @@ macro_rules! impl_diesel {
             // #[derive(diesel::AsExpression)]
             impl<Db> ToSql<Nullable<Binary>, Db> for $t
             where
-                for<'c> Db: Backend<BindCollector<'c> = RawBytesBindCollector<Db>>,
+                Db: Backend,
+                [u8]: ToSql<Binary, Db>,
```
mattsse, line 789: does

```
Self: ToSql<Binary, Db>,
```

work here?

at least this is what ruint does

https://github.com/recmo/uint/blob/ed6f5fc9f8b1369a28437a7b91cf224b8dbeef53/src/support/diesel.rs#L89C5-L89C29

crates/primitives/src/bits/macros.rs
```diff
@@ -757,14 +757,14 @@ macro_rules! impl_diesel {
...
-                query_builder::bind_collector::RawBytesBindCollector,
                 serialize::{Output, Result as SerResult, ToSql},
                 sql_types::{Binary, Nullable, SingleValue},
             };
 
             impl<Db> ToSql<Binary, Db> for $t
             where
-                for<'c> Db: Backend<BindCollector<'c> = RawBytesBindCollector<Db>>,
+                Db: Backend,
+                [u8]: ToSql<Binary, Db>,
```
mattsse, line 767: does this mean, this here is also wrong then?

https://github.com/recmo/uint/blob/ed6f5fc9f8b1369a28437a7b91cf224b8dbeef53/src/support/diesel.rs#L31

xavierm02: It's a bit harder for `Uint` because the bytes are stored internally as `[u64; LIMBS]`, and `diesel` doesn't provide `[u64]: ToSql<Binary, Db>` (and if you copy it into an `[u8]` to then pass that `[u8]` by reference, you don't get the right lifetime). I'm not sure what the right approach is in this case, the options being:
- using `bytemuck` to cast the reference and preserve its lifetime; or
- [implementing it for Sqlite independently](https://discord.com/channels/273534239310479360/1120124565591425034/1390695192218308690).
Both options could a priori bring endianness problems so one would have to pick carefully.

xavierm02:
Setting up proper test would be fairly tedious but here's a small repo that checks that it fixes insertion: https://github.com/xavierm02/test/tree/9f94495cc80c7a60e7d12bf0e73cfbc4bd5ebbb6/diesel

crates/primitives/src/bits/macros.rs
xavierm02: I'm looked a bit into it and I think implementing for Sqlite independently with `to_be_bytes_vec` should be fine. I'll do that on my next free evening.

xavierm02: I've looked a bit more into this, and the problem is that the only ways to implement `diesel`'s `ToSql` is to:
- either use an existing implementation of `ToSql`, but then you have to produce a reference with the same lifetime as the one you got, so this works if you send e.g. a field of a struct or of you can cast the reference, but if you have to transform the input, you can't send a reference to it because its lifetime would not be long enough,
- use backend-specific functions that allow you to pass an owned value.
The first solution is not possible for `ruint`, as the cast would only work for one endianness, and the second would mean either making `ruint` depend on `diesel` with all backend feature-flags activated, or adding a bunch of feature flags for the backends to `ruint`, neither of which is satisfactory.
The nicest way to do things I can see would be to add an `OwnedToSql` trait to `diesel`, but it's not clear whether that's something maintainers would accept, and it's probably a lot of work.

crates/primitives/src/bits/macros.rs
```diff
@@ -785,10 +785,11 @@ macro_rules! impl_diesel {
             // #[derive(diesel::AsExpression)]
             impl<Db> ToSql<Nullable<Binary>, Db> for $t
```
DaniPopes, line 786: `ToSql<Nullable<Binary` is not implemented for FixedBytes but it is for the macro-generated types, is this intended?

xavierm02: It is implemented by `#[cfg_attr(feature = "diesel", derive(diesel::AsExpression, diesel::FromSqlRow))]` for `FixedBytes`. The comment above this line says that this impl for `$t` is copied from what would be generated by the `AsExpression` derive macro. Not entirely sure why we want to copy the output of this proc macro instead of just using it directly in `wrap_fixed_bytes` though

xavierm02:
@DaniPopes @mattsse Is there something I can do to speed up the merge process?

DaniPopes (approved):

DaniPopes:
Thanks! Sorry for the long wait

PR #979, opened 2025-07-13
Add environment variable checks for WORKSPACE_ROOT and CRATE_ROOT in changelog script

Galoretka:
This PR adds explicit checks at the beginning of scripts/changelog.sh to ensure that the environment variables WORKSPACE_ROOT and CRATE_ROOT are set before proceeding. If either variable is unset, the script will exit with a clear error message. This improves the script's robustness and helps prevent accidental misconfiguration or unexpected behavior when running the changelog generation process.

### DIFF
--- a/scripts/changelog.sh
+++ b/scripts/changelog.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


DaniPopes:
this is not necessary

PR #983, opened 2025-07-25
Implement conversion from `Word` for `DynSolValue`

zemse:
## Motivation

```rs
let uint = U256::ZERO;
let address = Address::ZERO;
let word = B256::ZERO;

input(uint); // works
input(address); // works
input(word); // the trait bound `dynamic::value::DynSolValue: std::convert::From<alloy_primitives::FixedBytes<32>>` is not satisfied

fn input(v: impl Into<DynSolValue>) {}
```

## Solution

Implemented conversion from `Word` for `DynSolValue`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/dyn-abi/src/dynamic/value.rs
+++ b/crates/dyn-abi/src/dynamic/value.rs
@@ -117,6 +117,13 @@ impl From<Vec<u8>> for DynSolValue {
     }
 }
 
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
Thanks!

PR #993, opened 2025-08-22
feat(primitives): add borsh support

### LINKED ISSUE #992: [Feature] Add borsh support
### Component

primitives

following

https://github.com/recmo/uint/blob/25e81ad8c6732624ce92e7d12800715e78df726c/src/support/borsh.rs#L4

we can add this to the primitive types we introduce here, like fixedbytes

### Additional context

_No response_

mablr:
## Motivation
Close #992 

## Solution

- Added borsh as a dependency in Cargo.toml files for the main crate and primitives.
- Implemented borsh serialization and deserialization macros in the primitives library.
- Implemented borsh serialization and deserialization for `FixedBytes`

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -160,6 +160,7 @@ serde = [
 ]
 borsh = [
     "dep:borsh",
+    "ruint/borsh",
 ]
 
 allocative = ["dep:allocative"]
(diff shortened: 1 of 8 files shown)


mablr (commented):
Should we also implement `Bytes` ?

crates/primitives/Cargo.toml
```diff
@@ -155,6 +158,9 @@ serde = [
     "indexmap?/serde",
     "rand?/serde",
 ]
+borsh = [
+    "dep:borsh",
+]
```
mattsse, line 163: I think we should also activate in ruint/borsh here?

mattsse (changes requested):
cool, ty

one nit re feature activation

pending @DaniPopes

mattsse:
> Should we also implement Bytes

ah yes we should actually

PR #1003, opened 2025-09-02
feat(sol-macro-expander): add `Clone` trait to enum contracts containers

fd0r:
## Motivation

The `enum` containers associated to contracts function calls, custom errors and events, don't implement the `Clone` trait while the underlying `struct`s do.
Having the `Clone` trait propagated at this higher level is generally useful.

## Solution

Add `#[derive(Clone)]` where the containers are defined.

## PR Checklist

- [ ] Added Tests
    - No pre-existing test on which traits are implemented.
- [ ] Added Documentation
    - No pre-existing documentation on which traits are implemented in the custom containers.
- [ ] Breaking changes
    - Not a breaking change


mattsse (approved):
this should always work because we do the same for the variants

PR #1010, opened 2025-09-17
chore: fix syntax iss in the parser

famouswizard:
## Motivation

There were minor syntax issues in the code causing parsing errors:
1. A square bracket `[` was left unclosed.
2. The `|` immediately after `,` is invalid syntax.

Fixing these makes the code parse correctly and prevents runtime errors.

## Solution

Closed the unclosed square bracket and removed the invalid `|` after `,`.
No functional changes beyond correcting the syntax.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (changes requested):
can you run `cargo +nightly fmt --all` real quick? then this lgtm

Issue #999, opened 2025-08-29
[Feature] Move alloc feature behind a feature gate


Pzixel:
### Component

primitives

### Describe the feature you would like

I've checked the codebase, and two types that are used from the `alloc` crate are `String` and `vec`.

There is a possibility to move remove the `alloc` dependency from the crate. The only type that will be disabled by that is `Bytes` - but all Uxxx and Ixxx math, Address type and others will remain. 

There is a bunch of formatting that use String here and there, but it all can be replaced with on-stack buffers (stack-based `arrayvec::ArrayString` can be used instead), because for all those cases we either know the exact buffer size of at least the maximum possible string size. 

I think is reasonable to assume that `primitive_types` can work without the `alloc`, hence this issue. I can work on an implementation if this is something you consider to be accepted.

DaniPopes:
I think that's fair but only for alloy-primitives, the other crates make heavy use of these types

Pzixel:
Right, I'm talking about the `primitives` only.

SWASTIC-7:
@Pzixel , I am changing the String with fixed size ```pub type String = ArrayString<80>;```, There is lot of usage of `Vec` which are of dynamic size, do they also need to be changed to `ArrayVec` as in many case we don't know the size at compile time I guess? We are also using `Result<Box<MaybeUninit<T>>, TryReserveError>` which depends on `alloy` crate and i think we cannot completly get rid of the crate.

Pzixel:
@SWASTIC-7 Stack strings are only okay if you know the max length it will be. It's true for most of the types in the crate (like U256 for instance), but there are a lot of types that are inherently arbitrary sized. Then they can remain behind the feature gate, if they are naturally depending on this feature nothing you can do about it.

SWASTIC-7:
@Pzixel ok then I will make pr by changing the strings to stacked string wherever possible

PR #1013, opened 2025-09-19
Feat(primitive): moved some alloc:string to arrayvec::arrayString

SWASTIC-7:
implementing #999

## Motivation

The issue was to move String from `alloc::string::String` to `arrayvec::arrayString`

## Solution

Changed String to fixed capacity stack based String `pub type String = ArrayString<80>;`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -29,6 +29,7 @@ bytes.workspace = true
 hex.workspace = true
 itoa.workspace = true
 ruint.workspace = true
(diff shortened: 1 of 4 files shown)


DaniPopes:
this is not correct, Signed is fixed size but not to 80. also this is not what the issue is requesting, we would want to add a "alloc" feature and move users under that feature instead of using array strings.

PR #1015, opened 2025-09-21
chore(sol-types): sync panic reasons from geth

jsvisa:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-types/src/types/error.rs
+++ b/crates/sol-types/src/types/error.rs
@@ -419,7 +419,7 @@ impl PanicKind {
     /// Returns the panic code's string representation.
     pub const fn as_str(self) -> &'static str {
         // modified from the original Solidity comments:
-        // https://github.com/ethereum/solidity/blob/9eaa5cebdb1458457135097efdca1a3573af17c8/libsolutil/ErrorCodes.h#L25-L37
+        // <https://github.com/ethereum/solidity/blob/9eaa5cebdb1458457135097efdca1a3573af17c8/libsolutil/ErrorCodes.h#L25-L37>
         match self {
             Self::Generic => "generic/unspecified error",
             Self::Assert => "assertion failed",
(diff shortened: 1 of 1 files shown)


DaniPopes:
can we keep the default implementation of Display and instead have a separate method for getting the geth string?

jsvisa:
> can we keep the default implementation of Display and instead have a separate method for getting the geth string?

reverted back and add `as_geth_string`

crates/sol-types/src/types/error.rs
```diff
@@ -409,19 +421,19 @@ impl PanicKind {
 
     /// Returns the panic code's string representation.
     pub const fn as_str(self) -> &'static str {
-        // modified from the original Solidity comments:
-        // https://github.com/ethereum/solidity/blob/9eaa5cebdb1458457135097efdca1a3573af17c8/libsolutil/ErrorCodes.h#L25-L37
+        // Copy from go-ethereum's abi package:
+        // https://github.com/ethereum/go-ethereum/blob/4414e2833f92f437d0a68b53ed95ac5756a90a16/accounts/abi/abi.go#L261-L272
         match self {
```
DaniPopes, line 426: this should also be a separate function

crates/sol-types/src/types/error.rs
```diff
@@ -321,6 +321,18 @@ impl Panic {
         // use try_from to avoid copying by using the `&` impl
         u32::try_from(&self.code).ok().and_then(PanicKind::from_number)
     }
+
+    /// Returns the geth-compatible panic message string.
+    ///
+    /// This returns just the panic reason without the "panic:" prefix or hex code,
+    /// matching the format used by go-ethereum.
+    pub fn as_geth_string(&self) -> String {
```
DaniPopes, line 329: can be Cow<'static str>

mattsse (approved):

DaniPopes (approved):
pls fix ci

PR #1017, opened 2025-09-25
Fix internal SC derives

squadgazzz:
#1011 PR doesn't cover a case where SC bindings are generated from an ABI json, where the root SC depends on other interfaces that also generated from the same file. As a result, the compilation fails. This PR fixes that.

### DIFF
--- a/crates/sol-macro-input/src/json.rs
+++ b/crates/sol-macro-input/src/json.rs
@@ -27,6 +27,18 @@ impl SolInput {
         let derives =
             attrs.iter().filter(|attr| attr.path().is_ident("derive")).collect::<Vec<_>>();
 
+        // Extract all_derives and extra_derives attributes for propagation to internal contracts
+        let sol_derives = attrs
+            .iter()
+            .filter(|attr| attr.path().is_ident("sol"))
+            .filter_map(|attr| {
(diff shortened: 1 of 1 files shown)


squadgazzz:
I was able to reproduce it locally using this abi:
https://github.com/cowprotocol/services/blob/main/crates/contracts/artifacts/BalancerV2Vault.json
But for some reason, this test still passes with the abi above even without my fix:
https://github.com/alloy-rs/core/blob/40047491a702b7e06a131c992a2e27e544f65409/crates/sol-types/tests/macros/sol/json.rs#L219-L241

I am still trying to understand why I can't reproduce it in the alloy repo.

crates/sol-macro-input/src/json.rs
```diff
@@ -27,6 +27,27 @@ impl SolInput {
         let derives =
             attrs.iter().filter(|attr| attr.path().is_ident("derive")).collect::<Vec<_>>();
 
+        // Extract all_derives and extra_derives attributes for propagation to internal contracts
+        let sol_derives = attrs
+            .iter()
+            .filter(|attr| {
```
DaniPopes, line 33: you can extract this logic to a separate fn and merge it above into the derives filter

PR #1020, opened 2025-09-29
feat: Add Sqlx Traits for `Bytes` Type 

Arvmor:
## Motivation
This is the addition to #1008 PR

The `FixedBytes` a.k.a Address, B256, ... types already do impl sqlx Traits which enables crate users to directly pass the said types into sqlx queries. (and also Signed/UnSigned types)

Once I tried to do the same for `Bytes` type I realized that these traits are missing in sqlx crate feature, even though the underlying type is the same, so that motivated me to impl these traits.

## Solution
impl Type, Encode, Decode traits under the sqlx feature that is already existing.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/sqlx.rs
+++ b/crates/primitives/src/sqlx.rs
@@ -16,7 +16,7 @@ use sqlx_core::{
     types::Type,
 };
 
-use crate::{FixedBytes, Signed};
+use crate::{Bytes, FixedBytes, Signed};
 
 impl<const BYTES: usize, DB> Type<DB> for FixedBytes<BYTES>
 where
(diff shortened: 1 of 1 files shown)


crates/primitives/src/sqlx.rs
```diff
@@ -90,3 +90,37 @@ where
...
+
+impl<'a, DB: Database> Encode<'a, DB> for Bytes
+where
+    Vec<u8>: Encode<'a, DB>,
+{
+    fn encode_by_ref(
+        &self,
+        buf: &mut <DB as Database>::ArgumentBuffer<'a>,
+    ) -> Result<IsNull, BoxDynError> {
+        self.to_vec().encode_by_ref(buf)
```
mattsse, line 115: hmm, do we need the to_vec here or can we use the &[u8] impl for this?

mattsse (approved):
I guess we have the same for

https://github.com/alloy-rs/core/blob/e122aeb365e1dceadcbcea41dde7db8f2e67c827/crates/primitives/src/sqlx.rs#L35-L44

this this lgtm

pending @DaniPopes

DaniPopes (approved):

Issue #1024, opened 2025-10-02
[Feature] Handle `transient` keyword in `alloy_sol_macro`


AurelienFT:
### Component

sol! macro

### Describe the feature you would like

Hello,

Thanks for the work on alloy.
I tried to use the `alloy_sol_macro` (maybe it's an other crates used by the macro) and had issues with a contract using `transient` keyword.

Here is the line : 
```
    bool private transient locked;
 ```
Causing this error : 
```
expected a valid JSON ABI string or Solidity string: expected `;`
```

If it's because this keyword isn't handled in the parsing I'm willing to make a PR !

PR #1026, opened 2025-10-05
feat(sol-macro): add transient storage keyword support

AurelienFT:
Closes https://github.com/alloy-rs/core/issues/1024

## Motivation

Allow `alloy_sol_macro` to parse `transient` keyword that has been introduced in `0.28.8`.

## Solution

Add transient as a keyword in the storage sub-category

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

Side question: Is the Yul parsing still maintained ? If yes, is it ok to add `tstore` and `tload` opcodes ?


DaniPopes (approved):
Thanks!

PR #1041, opened 2025-11-17
feat(primitives): add `Signature::as_rsy` method

### LINKED ISSUE #1040: [Feature] Add helper to obtain normalized signature bytes
### Component

sol-types

see: https://github.com/alloy-rs/alloy/pull/3192

## TODO

* add another helper fn like `as_bytes` that passes in yparity byte as is

### Additional context

_No response_

mablr:
## Motivation

Close #1040

## Solution
add `Signature::as_rsy` method to obtain raw signature bytes, with y_parity as it is

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -165,12 +165,14 @@ impl Signature {
         sig
     }
 
-    /// Returns the byte-array representation of this signature with raw `y_parity`.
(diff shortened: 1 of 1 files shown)


mattsse:
hmm, actually, this isn't eip-2098, becasue we have that already and that is different

PR #1074, opened 2026-02-13
feat: move alloy-serde to primitives

stevencartavia:
## Motivation
moving `alloy-serde` to `alloy-primitives`

## Solution
Moved `alloy-serde` into `alloy-primitives` as a `pub mod serde` behind the existing serde feature flag. Source files are unchanged aside from import path adjustments (alloy_primitives:: → crate::) and feature gate fixes needed because `arbitrary` and `map` are optional in primitives. Added `serde_json` as an optional dependency gated behind the `serde` feature.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


mattsse (commented):
all of this could we some additional bloat, however imo this is okay, but maybe it's worth gating the serde-json support separately, like serde-json feature

wdyt @DaniPopes

PR #1090, opened 2026-03-24
perf: add Token::head_append_many for batch encoding

DaniPopes:
Adds a `head_append_many` method to the `Token` trait for bulk encoding a slice of tokens. The default implementation loops over `head_append`.

`WordToken` overrides this to reinterpret `&[WordToken]` as `&[Word]` (both `#[repr(transparent)]`) and calls `Encoder::append_words` (`extend_from_slice`), replacing N individual `Vec::push` calls with a single memcpy.

`FixedSeqToken` and `DynSeqToken` use `head_append_many` in their `head_append` and `encode_sequence` for non-dynamic inner tokens, also skipping the unnecessary offset tracking in that case.

### DIFF
--- a/crates/sol-types/src/abi/token.rs
+++ b/crates/sol-types/src/abi/token.rs
@@ -359,21 +359,7 @@ impl<'de, T: Token<'de>, const N: usize> Token<'de> for FixedSeqToken<T, N> {
 
 impl<'de, T: Token<'de>, const N: usize> TokenSeq<'de> for FixedSeqToken<T, N> {
     fn encode_sequence(&self, enc: &mut Encoder) {
-        if T::DYNAMIC {
-            enc.push_offset(self.0.iter().map(T::head_words).sum());
-
-            for inner in &self.0 {
-                inner.head_append(enc);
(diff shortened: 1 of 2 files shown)


mattsse (approved):

crates/sol-types/src/abi/token.rs
```diff
@@ -565,6 +567,26 @@ impl PackedSeqToken<'_> {
     }
 }
 
+/// Shared implementation for [`TokenSeq::encode_sequence`] used by both
+/// [`FixedSeqToken`] and [`DynSeqToken`].
+fn encode_sequence_impl<'de, T: Token<'de>>(tokens: &[T], enc: &mut Encoder) {
+    if T::DYNAMIC {
```
DaniPopes: if !T::DYNAMIC, `push_offset` pushes 0, `bump_offset` increases it by `inner.tail_words` which is 0, `tail_append` is a noop, and `pop_offset` pops the 0 off the stack. so this whole thing reduces to `for x in tokens { x.head_append(enc) }` -- then this optimizes to head_append_many.

PR #1098, opened 2026-05-06
chore(deps): bump all

DaniPopes:
new cryptoes and hashbrown. pinning bincode since auto update goes to 3.0 and it's borked.


mattsse (approved):
bumpies

looks like we need to update a test

PR #1102, opened 2026-05-13
fix(EIP-712): PropDef::parse supports extra whitespaces

anaPerezGhiglia:
## Motivation

Current implementation fails to parse EIP-712 types containing certain whitespace: specifically, whitespace immediately before a `)` or `,`. Leading whitespace after `(` and after `,` are already tolerated. 

The root cause is `PropDef::parse` relying on `rsplit_once(' ')` without trimming its input first. As a result, in the cases mentioned above, the split leaves an internal space inside the type half that `TypeSpecifier::parse_eip712` then rejects.

## Solution

Trim `PropDef::parse` `input` parameter before splitting and parsing it.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
Thanks!

PR #1104, opened 2026-05-14
feat: add optional `secp256k1` backend for ECDSA operations

decofe:
## Summary

Add a `secp256k1` feature (disabled by default) to `alloy-primitives` and `alloy-core` that uses the C libsecp256k1 bindings ([`secp256k1`](https://crates.io/crates/secp256k1) v0.31) as an alternative to the pure-Rust `k256` crate for elliptic curve operations.

When both `k256` and `secp256k1` features are enabled, **`secp256k1` is preferred** for address recovery operations.

### Feature wiring
- Workspace dep: `secp256k1 = { version = "0.31", default-features = false }`
- `alloy-primitives`: optional dep with `alloc` + `recovery` features; `global-context` enabled under `std`
- `alloy-core`: forwards `secp256k1` feature to `alloy-primitives`

### Backend-neutral recovery
`recover_address_from_msg` and `recover_address_from_prehash` now compile under `#[cfg(any(feature = "k256", feature = "secp256k1"))]` and use `cfg_if!` to select the preferred backend:
1. `secp256k1` (if enabled) — uses `Secp256k1::recover_ecdsa` + `RecoverableSignature`
2. `k256` (fallback) — uses `VerifyingKey::recover_from_prehash`

### New secp256k1-specific API surface
- `Signature::to_secp256k1()` / `to_secp256k1_recoverable()`
- `Signature::from_secp256k1_signature_and_parity()`
- `Signature::recover_secp256k1_from_prehash()`
- `Address::from_secp256k1_public_key()` / `from_secp256k1_secret_key()`
- `From<(RecoverableSignature, RecoveryId)> for Signature`
- `TryFrom<Signature> for secp256k1::ecdsa::Signature`
- `SignatureError::Secp256k1` variant

### Preserved
All existing `k256`-gated API is untouched — no breaking changes.

## Testing
Tested with all feature combinations:
- `--features secp256k1` ✅
- `--features k256` ✅ 
- `--features k256,secp256k1` ✅
- `--no-default-features --features secp256k1` ✅

Recovery tests (`can_recover_tx_sender_not_normalized`, `recover_web3_signature`) broadened to run under both backends.

### DIFF
--- a/crates/primitives/src/signature/sig.rs
+++ b/crates/primitives/src/signature/sig.rs
@@ -4,7 +4,7 @@ use crate::{B256, U256, hex, normalize_v, signature::SignatureError, uint};
 use alloc::vec::Vec;
 use core::{fmt::Display, str::FromStr};
 
-#[cfg(feature = "k256")]
+#[cfg(any(feature = "k256", feature = "secp256k1"))]
 use crate::Address;
 
 /// The order of the [Secp256k1](https://en.bitcoin.it/wiki/Secp256k1) curve.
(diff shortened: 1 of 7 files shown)


DaniPopes (approved):

crates/primitives/src/signature/sig.rs
```diff
@@ -90,6 +90,27 @@ impl TryFrom<Signature> for k256::ecdsa::Signature {
     }
 }
 
+#[cfg(feature = "secp256k1")]
+impl From<(secp256k1::ecdsa::RecoverableSignature, secp256k1::ecdsa::RecoveryId)> for Signature {
+    fn from(value: (secp256k1::ecdsa::RecoverableSignature, secp256k1::ecdsa::RecoveryId)) -> Self {
+        let (_, compact) = value.0.serialize_compact();
```
mattsse: this drops the `RecoveryId` here, replacing it with the one from the tuple, is that intended?

mattsse (approved):

PR #1105, opened 2026-05-14
fix(dyn-abi): allow self-referential struct types in EIP-712 canonicalize (#1103)

### LINKED ISSUE #1103: [Bug] `EncodeType::canonicalize` rejects EIP-712 recursive struct types
### Component

dyn-abi

`alloy-dyn-abi 1.5.7` (reproduced on `alloy-rs/core@00ee366`).

### Operating System

None

### Describe the bug

`EncodeType::canonicalize` rejects EIP-712 type definitions that contain recursion, even though the EIP-712 spec explicitly supports them. The simplest repro is a self-referential struct.

## Spec

From [EIP-712](https://eips.ethereum.org/EIPS/eip-712):

> The standard supports recursive struct types.

And from the Rationale:

> The current standard is optimized for tree-like data structures and undefined for cyclical data structures.

cyclical *data instances* (runtime values that loop back on themselves) are undefined, but cyclical *type definitions* are allowed: they describe possibly-recursive shapes, and concrete instances must still be finite trees.

## Repro

The following unit test, added to `crates/dyn-abi/src/eip712/parser.rs`, fails on `main`:

```rust
#[test]
fn canonicalize_self_referential_type() {
    // Per EIP-712: "The standard supports recursive struct types."
    let input = "Node(uint256 value,Node[] children)";
    let encoded = EncodeType::parse(input).unwrap();
    assert_eq!(encoded.canonicalize(), Ok(input.to_string()));
}
```

### Actual output

```
thread 'eip712::parser::tests::canonicalize_self_referential_type' panicked at crates/dyn-abi/src/eip712/parser.rs:...
assertion `left == right` failed
  left: Err(MissingType("primary component"))
 right: Ok("Node(uint256 value,Node[] children)")
```

## Note

`EncodeType::parse` accepts the input without complaint. The parser/grammar already considers recursive type definitions well-formed. The rejection happens in `canonicalize`, which relies on `non_dependent_types` returning at least one type to identify the primary component. A self-referential type is treated as "dependent" on itself, so the iterator comes back empty and `canonicalize` errors out with `MissingType("primary component")` before any cycle check is reached.

0xSoftBoi:
## Summary

- `EncodeType::canonicalize` rejected recursive struct types such as `Node(uint256 value, Node[] children)` with `MissingType("primary component")`, even though [EIP-712] states *"The standard supports recursive struct types."*
- `Resolver::non_dependent_types` treated a self-edge as making a type dependent on itself, so the primary component iterator came back empty and `canonicalize` errored out before any cycle check.

## Fix

- `non_dependent_types` now ignores self-edges when computing the dependent set, so a self-referential type is still recognised as a non-dependent (primary) component.
- Cycle detection is split into a strict variant (used by `Resolver::resolve`, which builds a `DynSolType` and therefore cannot represent recursive types) and a permissive variant used by `Resolver::linearize` / `Resolver::encode_type`. Cycles between distinct types are still rejected on both paths.
- `canonicalize` no longer calls the redundant `resolver.resolve(primary)`; `encode_type` already walks the dependency graph via `linearize` and surfaces missing-type and circular-dependency errors.

`TypedData::eip712_signing_hash` still returns `CircularDependency` for self-referential types because the `encode_data` path goes through `resolve`; only the `encodeType` / `typeHash` path becomes permissive. The existing `test_hash_recursive_types` continues to pass unchanged.

## How it was tested

New tests (all pass; `cargo fmt` and `cargo clippy -- -D warnings` are clean):

- `eip712::parser::tests::canonicalize_self_referential_type` — the exact repro from the issue.
- `eip712::parser::tests::canonicalize_self_referential_type_with_dependency` — recursive primary with an extra leaf type.
- `eip712::resolver::tests::non_dependent_self_referential_type_is_primary`
- `eip712::resolver::tests::encode_type_round_trip_self_referential`
- `eip712::resolver::tests::resolve_still_rejects_self_referential_types`
- `eip712::resolver::tests::encode_type_still_rejects_mutual_recursion`

Pre-existing `test_fails_encode_type_circular` (mutual recursion through distinct types) and `test_hash_recursive_types` continue to pass.

Full `cargo test -p alloy-dyn-abi --all-features`: **194 passed; 0 failed.**

Fixes #1103

[EIP-712]: https://eips.ethereum.org/EIPS/eip-712

### DIFF
--- a/crates/dyn-abi/src/eip712/parser.rs
+++ b/crates/dyn-abi/src/eip712/parser.rs
@@ -184,9 +184,13 @@ impl<'a> EncodeType<'a> {
         };
 
         let primary = first.type_name();
-        _ = resolver.resolve(primary)?;
 
-        // Encode primary type
+        // Encode primary type. This delegates to `Resolver::encode_type`, which
+        // walks the dependency graph via `Resolver::linearize` and surfaces both
+        // missing-type and circular-dependency errors. We intentionally do not
+        // call `Resolver::resolve` here, because that builds a `DynSolType` and
+        // therefore cannot represent recursive struct types (which EIP-712
+        // explicitly permits, see <https://eips.ethereum.org/EIPS/eip-712>).
         resolver.encode_type(primary)
     }
 }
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):

0xSoftBoi:
Friendly ping — this is approved and mergeable. Happy to rebase if needed, otherwise let me know if anything else is required before merge. Thanks!

PR #1109, opened 2026-05-23
perf(sol-types): avoid allocating event signature errors

mattsse:
## Summary

Adds a dedicated `Error::InvalidEventSignatureHash` variant that stores the event signature name plus the received and expected hashes directly.

## Why

`Error::invalid_event_signature_hash` previously built an owned formatted string immediately, forcing an allocation even when callers only propagate the error. The new variant keeps construction allocation-free and defers formatting to `Display`.

## Impact

Event signature validation errors now retain structured mismatch data and preserve the existing display message without allocating during error construction.

### DIFF
--- a/crates/sol-types/src/errors.rs
+++ b/crates/sol-types/src/errors.rs
@@ -51,6 +51,16 @@ pub enum Error {
         max: u8,
     },
 
+    /// Invalid event signature hash.
+    InvalidEventSignatureHash {
+        /// The event signature.
+        name: &'static str,
+        /// The received signature hash.
+        got: B256,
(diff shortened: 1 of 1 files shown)


crates/sol-types/src/errors.rs
```diff
@@ -178,3 +196,34 @@ impl From<TryReserveError> for Error {
         Self::Reserve(value)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
DaniPopes: we dont need a slop test for this please

DaniPopes (approved):

PR #1122, opened 2026-06-10
fix(dyn-abi): validate event topic count before decoding body

satyakwok:
Fixes alloy-rs/core#1123.

## Problem

`DynSolEvent::decode_log_parts` decoded the body data **before** fully validating the topic count. The up-front `size_hint` check only fires for exact-size iterators, so a log whose topic count does not match the event — when the topics are supplied via a non-exact iterator (`filter`/`map`/streaming) — reached the body decode and failed with a confusing `Overrun` instead of `TopicLengthMismatch`.

This shows up in practice because ERC-20 and ERC-721 `Transfer` share the same `topic0` (`keccak("Transfer(address,address,uint256)")`), so a topic-only filter matches both. Decoding an ERC-721 transfer log (4 topics, empty data) against the ERC-20 `Transfer` ABI displaced the body words and overran:

```text
exact   iterator -> TopicLengthMismatch { expected: 3, actual: 4 }   // ok
inexact iterator -> SolTypes(Overrun)                                // the bug
```

## Fix

Decode the indexed topics, validate the topic count, **then** decode the body. Running out of topics now also reports `TopicLengthMismatch` rather than silently producing a short indexed list. Valid logs are unaffected; the early `size_hint` fast-path is kept.

No public API or error-variant changes (`TopicLengthMismatch` already exists).

## Tests

- `topic_count_is_validated_before_body` — over-count via both exact and non-exact iterators (the reported bug).
- `too_few_topics_is_a_length_mismatch` — under-count via a non-exact iterator.
- Valid-log decoding is covered by the existing `it_decodes_logs_with_indexed_params`.

`cargo test -p alloy-dyn-abi` → 160 passed; `cargo fmt --check` clean; `clippy --all-features` with `-D warnings` clean.

### DIFF
--- a/crates/dyn-abi/src/dynamic/event.rs
+++ b/crates/dyn-abi/src/dynamic/event.rs
@@ -276,7 +276,8 @@ mod test {
         };
 
         // Only sig + one indexed topic, supplied via a non-exact iterator.
-        let topics = [t0, b256!("0x000000000000000000000000d9a442856c234a39a81a089c06451ebaa4306a72")];
+        let topics =
+            [t0, b256!("0x000000000000000000000000d9a442856c234a39a81a089c06451ebaa4306a72")];
         let err = event.decode_log_parts(topics.iter().copied().filter(|_| true), &[]).unwrap_err();
         assert!(matches!(err, Error::TopicLengthMismatch { expected: 3, actual: 2 }), "{err:?}");
     }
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

satyakwok:
Thanks for the review and the clippy fix — should've run clippy with `--all-targets` to catch that one. Appreciated!

PR #1129, opened 2026-06-14
fix(sol-macro): derive Default for dynamic arrays of non-Default elements

satyakwok:
A dynamic Solidity array maps to `Vec<T>`, whose `Default` impl holds for any `T`. `can_derive_default` still required the element type to be `Default` for dynamic arrays, so a struct/error with a field like `MyEnum[]` lost its `Default` derive under `#[sol(all_derives)]` even though `Vec<MyEnum>: Default`.

This mirrors how the field's Rust type is generated (`expand_rust_type_to`): `Some(size)` → `[T; N]`, `None` → `Vec<T>`. Only fixed-size arrays need `T: Default` (and `N <= 32`); dynamic arrays are always `Default`.

Added a regression test (`test_default_dynamic_array_of_non_default_elem`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,8 @@ jobs:
           persist-credentials: false
       - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
         with:
-          toolchain: nightly
+          # Pin docs to a known-good nightly until the 2026-07-02 rustdoc ICE is fixed.
+          toolchain: nightly-2026-06-27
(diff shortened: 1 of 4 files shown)


satyakwok:
@DaniPopes,  @mattsse  small follow-up near the same area as #1124 — `can_derive_default` required the element to be `Default` even for dynamic arrays, but `T[]` is `Vec<T>` which is always `Default` (only fixed `[T; N]` needs `T: Default`). Mirrors the `eval_array_size` predicate from `expand_rust_type_to`. Mind taking a look when you get a chance?

DaniPopes (approved):

PR #1140, opened 2026-07-01
Rename unchecked sol decode APIs

decofe:
## Summary
- rename ABI decode APIs that skip validation to use `_unchecked` suffixes
- update macro expansion, doctests, README examples, and generated-code tests
- rename `SolEvent::new` to `new_unchecked` to match `new_checked`

## Testing
- cargo check -p alloy-sol-types -p alloy-sol-macro-expander
- cargo test -p alloy-sol-types

Prompted by: @0xalpharush


DaniPopes:
breaking change, also the current default matches other ABI libraries; validation is adhoc and not really specified in any way

PR #1143, opened 2026-07-03
Fix dyn-abi custom struct proptest regression

decofe:
## Summary
- keep the existing dyn-abi decode behavior unchanged
- skip ABI round-trip equality when the expected type contains a custom struct, since decoding cannot repopulate custom struct names and ZST custom structs can disappear from the decoded value shape
- add regression seeds for the dynamic and fixed zero-sized custom struct cases found by proptest

## Tests
- cargo test -p alloy-dyn-abi --lib arbitrary::tests::arbitrary_value --all-features
- cargo test -p alloy-dyn-abi --lib --all-features
- git diff --check

Prompted by: @0xalpharush


decofe:
Closing in favor of #1142; the scoped proptest patch has been pushed to that PR branch.

PR #1146, opened 2026-07-05
fix(syn-solidity): strip namespaces from custom types in EIP-712 formatting

Ysh204:
## Motivation

EIP-712 type signatures should use the bare name of referenced custom types.

While direct custom types were formatted correctly, custom types nested inside
arrays retained their fully qualified paths (e.g. `A.StructB[]`), resulting in
incorrect EIP-712 type signatures and hashes.

## Solution

Add a private recursive helper for EIP-712 type formatting that strips
namespaces from custom types while recursively formatting array element types.

Also adds regression tests covering:

- Direct custom types
- Dynamic arrays
- Fixed-size arrays
- Nested arrays

to ensure custom type names are formatted correctly in all supported array
configurations.

### DIFF
--- a/tests/core-sol/src/lib.rs
+++ b/tests/core-sol/src/lib.rs
@@ -207,25 +207,3 @@ fn do_stuff() {
     assert_eq!(set.len(), 1);
     assert_eq!(MyType::NAME, "MyType");
 }
-
-sol! {
-    contract A {
-        struct StructB {
-            uint256 x;
-        }
-    }
-    struct StructA {
-        A.StructB[] b;
-        A.StructB[3] c;
-        A.StructB[][] d;
-        A.StructB e;
-    }
-}
-
-#[test]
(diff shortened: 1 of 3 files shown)


tests/core-sol/src/lib.rs
```diff
@@ -207,3 +207,25 @@ fn do_stuff() {
     assert_eq!(set.len(), 1);
     assert_eq!(MyType::NAME, "MyType");
 }
+
+sol! {
```
DaniPopes: wrong place for the test

Ysh204: I've moved the regression test to crates/sol-types/tests/macros/sol/eip712.rs alongside the other EIP-712 macro tests and verified that cargo test -p alloy-sol-types passes.

DaniPopes (approved):

PR #1147, opened 2026-07-12
chore(crates/primitives): migrate to new versions of `rand` and `k256`

StackOverflowExcept1on:
## Motivation

Since new version of [`k256`](https://docs.rs/k256) has been released, it would be wise to support it

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
         rust: [
             "stable",
             "nightly",
-            "1.88", # MSRV
+            "1.90", # MSRV
           ]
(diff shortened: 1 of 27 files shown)


mablr (commented):
@StackOverflowExcept1on could you please revive this PR ? fix fmt/deny

StackOverflowExcept1on:
@mablr I would like to know if `ruint` update is required first?

mattsse (commented):
once this is merged we also need to do the same for all the other alloy reps

PR #1148, opened 2026-07-12
fix(sol-macro): correct UDVT code generation for dynamic types

Ysh204:
## Motivation

The generated `SolTypeValue` implementation for user-defined value types (UDVTs) wrapping dynamically-sized Solidity types delegated several methods to the wrong APIs.

This caused:
- `stv_eip712_data_word` to use `SolType::tokenize`, which does not produce the expected `Word` for dynamically-sized types.
- `stv_abi_packed_encoded_size` to use `abi_encoded_size` instead of the packed size calculation, producing incorrect packed sizes for dynamic underlying types.
- `into_underlying` to be generated as a `const fn`, which is not valid for underlying types with destructors (e.g. `String`).

## Solution

Update the generated `SolTypeValue` implementation to delegate to the underlying `SolTypeValue` methods instead of the corresponding `SolType` methods.

Additionally:
- Remove the unnecessary `const` qualifier from `into_underlying`.
- Add a regression test covering a UDVT wrapping `string` to verify the generated type compiles successfully and reports the correct packed encoded size.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/sol-macro-expander/src/expand/udt.rs
+++ b/crates/sol-macro-expander/src/expand/udt.rs
@@ -48,17 +48,17 @@ pub(super) fn expand(cx: &ExpCtxt<'_>, udt: &ItemUdt) -> Result<TokenStream> {
 
                 #[inline]
                 fn stv_eip712_data_word(&self) -> alloy_sol_types::Word {
-                    <#underlying_sol as alloy_sol_types::SolType>::tokenize(self).0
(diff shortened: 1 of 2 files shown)


DaniPopes:
UDVTs cannot be dynamic

PR #1153, opened 2026-07-16
feat(primitives): use foldhash for fixed-size maps

onbjerg:
Fixed-size map aliases such as U256Map now use foldhash's fast RandomState through FbBuildHasher instead of FxHash. The fixed-byte map aliases retain their existing API and specialized hashing path, while foldhash supplies randomized per-builder state in both std and no_std builds. The old rustc-hash dependency is now only needed when map-fxhash is enabled.

Generic map::HashMap defaults remain unchanged. The fixed-map test suite, feature checks, and workspace compilation pass.

### DIFF
--- a/crates/primitives/src/map/fixed.rs
+++ b/crates/primitives/src/map/fixed.rs
@@ -52,12 +52,14 @@ macro_rules! fb_alias_maps {
 
 fb_alias_maps!(Selector<4>, Address<20>, B256<32>, U256<32>);
 
-type FbBuildHasherInner = super::FxBuildHasherInner;
-type FbHasherInner = rustc_hash::FxHasher;
+type FbBuildHasherInner = foldhash::fast::RandomState;
+type FbHasherInner = foldhash::fast::FoldHasher<'static>;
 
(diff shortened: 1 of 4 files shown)


crates/primitives/src/map/fixed.rs
```diff
@@ -52,12 +52,14 @@ macro_rules! fb_alias_maps {
...
-type FbBuildHasherInner = super::FxBuildHasherInner;
-type FbHasherInner = rustc_hash::FxHasher;
+type FbBuildHasherInner = foldhash::fast::RandomState;
+type FbHasherInner = foldhash::fast::FoldHasher<'static>;
 
 /// [`BuildHasher`] optimized for hashing [fixed-size byte arrays](FixedBytes).
 ///
 /// **NOTE:** this hasher accepts only `N`-length byte arrays! It is invalid to hash anything else.
+///
+/// The hasher uses the randomly initialized fast foldhash state.
```
DaniPopes: rm

DaniPopes (approved):

PR #1155, opened 2026-07-17
feat(json-abi): support enum definitions

stevencartavia:
JSON ABIs preserve an enum’s name through `internalType`, but do not include its variants, so `JsonAbi::to_sol` currently represents enums as `uint8` user-defined value types. This adds optional enum definitions to the JSON ABI rendering and `SolInput` normalization APIs, allowing callers with compiler metadata to generate real Solidity enum declarations.

Existing behavior remains unchanged when definitions are missing, invalid, or conflicting, falling back to the current `uint8` representation. This enables consumers such as Foundry to restore enum variants from Solar without duplicating Alloy’s ABI traversal or rewriting the generated AST. This supports [foundry-rs/foundry#10441](https://github.com/foundry-rs/foundry/issues/10441).


crates/json-abi/src/abi.rs
```diff
@@ -204,12 +204,37 @@ impl JsonAbi {
         out
     }
 
+    /// Formats this JSON ABI as a Solidity interface using known enum definitions.
+    #[inline]
+    pub fn to_sol_with_enums(
```
DaniPopes: please don't add new functions or params, that's the point of tosolconfig

stevencartavia: [679acf7](https://github.com/alloy-rs/core/pull/1155/commits/679acf79e7512ec3ae29720db40ee87a17fe0ddb)

crates/json-abi/src/to_sol.rs
```diff
@@ -13,6 +13,12 @@ use core::{
     ops::{Deref, DerefMut},
 };
 
+/// Solidity enum variants keyed by unqualified (`Status`) or qualified (`Token.Status`) names.
+///
+/// Variants must be in discriminant order. Missing, conflicting, or invalid definitions fall back
+/// to a `uint8` UDVT.
+pub type EnumDefinitions = BTreeMap<String, Vec<String>>;
```
DaniPopes: no need for this to be public

crates/json-abi/src/to_sol.rs
```diff
@@ -49,14 +57,22 @@ impl ToSolConfig {
...
+    /// Enums with a supplied definition are printed as enum declarations; otherwise they are
+    /// printed as `uint8` user-defined value types. If disabled, supplied definitions are ignored.
     #[inline]
     pub const fn enums_as_udvt(mut self, yes: bool) -> Self {
         self.enums_as_udvt = yes;
         self
     }
 
+    /// Sets supplemental enum definitions used when formatting the ABI.
+    pub fn enum_definitions(mut self, definitions: EnumDefinitions) -> Self {
```
DaniPopes: use impl IntoIterator

crates/json-abi/src/to_sol.rs
stevencartavia: [b712ece](https://github.com/alloy-rs/core/pull/1155/commits/b712ecedb1e1a92cf8b0da2767a6b611ae0c565c)

crates/json-abi/src/to_sol.rs
stevencartavia: [b712ece](https://github.com/alloy-rs/core/pull/1155/commits/b712ecedb1e1a92cf8b0da2767a6b611ae0c565c)

DaniPopes (approved):

PR #1164, opened 2026-08-18
fix(sol-types): validate dynamic array length before allocation

rappie:
## Motivation

Dynamic-array decoding allocated its output buffer before checking whether enough data was available, doing unnecessary work when decoding could not proceed.

## Solution

Check the available ABI words before allocation while preserving zero-sized element behavior. Add regression coverage for both paths.

Codex was used to help understand the existing code and assist with the implementation. I fully understand and manually reviewed the resulting changes.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


0xalpharush:
This fixes the reported non-ZST preallocation case, but I think two cases remain before the mitigation is complete:

1. **The ZST length-controlled loop remains.** When `size_of::<T>() == 0`, the preflight is skipped, but the default `T::decode_many_from` still iterates `len` times. A tiny `Array<()>` encoding declaring billions of elements therefore performs billions of decode iterations. The two-element regression test preserves behavior but does not cover the DoS fixed for dyn-abi in #1142. We should short-circuit `T::MINIMUM_WORDS == 0` to the existing empty ZST representation and add a `u32::MAX`-sized regression like #1142.

2. **One word per element undercounts static composite elements.** For example, an element of type `uint256[1024]` requires 1,024 words in the containing array head, but this check accepts one word before reserving the full Rust representation. This still permits allocation amplification relative to the physically available input.

I suggest making the footprint a sealed `Token` invariant and checking it before allocation:

```rust
if T::MINIMUM_WORDS == 0 {
    return Ok(Self(Vec::new()));
}

let required_words = T::MINIMUM_WORDS
    .checked_mul(len)
    .ok_or(Error::Overrun)?;
if required_words > child.remaining_words() {
    return Err(Error::Overrun);
}
```

`MINIMUM_WORDS` should mean the immediate containing-head width: `1` for a dynamic token, the summed/multiplied static width for a static tuple/fixed array, and `0` for ZSTs. This avoids counting dynamic tails, so it remains compatible with solc's array-head bounds check.

Useful additional tests would cover a multiword static element, checked-multiplication overflow, a dynamic element requiring one offset word, and a huge ZST length.

rappie:
Should be ready for review again.

decofe:
The zero-sized-element fast path changes decoding semantics: `Array<()>` with encoded length 2 now returns `[]` instead of `[(), ()]`.

The ABI specification defines `T[]` as the encoded length followed by the tuple of its `k` elements, and explicitly permits tuples with `k >= 0`; zero-width elements therefore consume no bytes but still retain their element count. Alloy supports this type directly, and ethers.js/ethabi likewise preserve the decoded count.

Could we avoid iterating over an attacker-controlled length while still constructing a zero-sized `Vec<T>` with `len` elements? The regression test should assert that a small nonzero length round-trips rather than expecting `vec![]`.

Spec: https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding

DaniPopes (approved):

PR #1165, opened 2026-08-20
fix(sol-types): limit ABI decoder memory

decofe:
## Motivation

Small ABI payloads can reuse offsets so the decoder visits the same encoded data many times and allocates a new value on every visit.


DaniPopes:
https://github.com/alloy-rs/core/pull/1167

PR #1175, opened 2026-08-27
feat(primitives): add Postgres support to fixed-bytes wrappers

### LINKED ISSUE #1032: [Feature] To/From Sql for address type in postgres mod
### Component

primitives

Are there any plans to add support for the address type into the postgres mod? It looks like there are implementations for diesel and sqlx, but my use case is for tokio postgres. Would be nice to have this supported directly, without having to utilize a wrapped type. New to rust, but I am happy to offer up what I am using.

### Additional context

_No response_

zexoverz:
Closes #1032. `Address` is declared with `wrap_fixed_bytes!` rather than being an alias, so the blanket `FixedBytes<BITS>` impls at the top of the module never applied to it and users had to wrap it themselves.

Kept to BYTEA only, matching what the `FixedBytes` pair accepts. `Signed` supports text and numeric types too but an address has no meaningful numeric reading, and adding a hex TEXT path would be a guess at what people store rather than something the issue asked for.

### DIFF
--- a/crates/primitives/src/bits/macros.rs
+++ b/crates/primitives/src/bits/macros.rs
@@ -235,6 +235,7 @@ macro_rules! wrap_fixed_bytes {
         $crate::impl_allocative!($name);
         $crate::impl_arbitrary!($name, $n);
         $crate::impl_rand!($name);
+        $crate::impl_postgres!($name, $n);
         $crate::impl_diesel!($name, $n);
         $crate::impl_sqlx!($name, $n);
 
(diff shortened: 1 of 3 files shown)


onbjerg (approved):

zexoverz:
The red is `allocative 0.3.6` failing to compile on nightly, where its `Infallible` and `!` impls now collide. `main` has been failing the same way since 1.7.1, so nothing here comes from this PR.

onbjerg:
yep, can't merge while this is failing though. will rebase once fixed.

DaniPopes (approved):

DaniPopes:
thanks, moved to the macro itself to implement it for all types

PR #1185, opened 2026-09-09
fix(json-abi): preserve arrays when printing enums as integers

sueun-dev:
## Motivation

With `ToSolConfig::enums_as_udvt(false)`, `JsonAbi::to_sol` prints an enum array such as `uint8[]` as scalar `uint8`. This changes the generated interface's parameter types, including nested struct fields, event inputs, and return values.

## Solution

Use the ABI type when replacing an enum name, preserving its array dimensions. Add a regression covering dynamic, fixed, and multidimensional arrays, with scalar enums and ordinary integer arrays as controls. Also check the array input in `known_enum_definitions` when enum printing is disabled.

The new regression fails before the fix. I compiled its fixture with solc 0.8.30 and recompiled the printed interface: the parameter and return types differ before the fix and match afterward.

Local checks passed on Linux aarch64 with Rust 1.94.1:

- `cargo test -p alloy-json-abi --all-features --offline`
- `cargo test -p alloy-json-abi --no-default-features --offline`
- `cargo test -p alloy-dyn-abi -p alloy-sol-types -p alloy-sol-macro --all-features --offline` (upstream-ignored tests remain ignored)

Formatting, package Clippy with warnings denied, and package documentation also passed using nightly-2026-06-27.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation (not needed for this bug fix)
- [ ] Breaking changes


DaniPopes:
this is not correct

PR #1186, opened 2026-09-10
feat(sol-types): add validation that allows trailing bytes

decofe:
## Motivation

Strict ABI decoding rejects a canonical encoding followed by extra bytes, while Solidity accepts trailing bytes. Callers need to accept that suffix while retaining Alloy's strict layout, padding, and token checks.

## Solution

Add `AbiDecoderConfig::validate_allow_trailing_bytes(true)`, with matching getter and in-place setter. It enables strict validation of the encoded prefix and skips only the final full-input-consumption check. The option permits trailing bytes even when `strict(true)` is also set, independent of setter order.

```rust
let config = AbiDecoderConfig::new().validate_allow_trailing_bytes(true);
let value = MyType::abi_decode_with_config(data, config)?;
```

This matches Solidity only with respect to trailing bytes. Canonical offsets, complete zero padding, UTF-8 validity, and other token checks remain enforced.

Tests cover empty, static, and nested dynamic values; single-value, parameter, and sequence decoding; aligned and unaligned suffixes; configuration interactions; and rejection of truncated data, dirty padding, gaps, overlaps, invalid bools, and invalid UTF-8.

## Validation

- `cargo test -p alloy-sol-types --all-features`
- `cargo +nightly clippy -p alloy-sol-types --all-targets --all-features`
- `cargo check -p alloy-sol-types --no-default-features`
- `cargo +nightly fmt --all -- --check`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

Prompted by: @klkvr

### DIFF
--- a/crates/sol-types/src/abi/decoder.rs
+++ b/crates/sol-types/src/abi/decoder.rs
@@ -102,7 +102,7 @@ impl AbiDecoderConfig {
     /// is enabled.
     #[inline]
     pub const fn get_strict(&self) -> bool {
-        self.strict || self.validate_allow_trailing_bytes
+        self.strict
     }
 
     /// Returns whether strict validation permits trailing bytes.
--- a/crates/sol-types/src/abi/decoder.rs
+++ b/crates/sol-types/src/abi/decoder.rs
@@ -150,13 +150,14 @@ impl AbiDecoderConfig {
         self
     }
 
-    /// Enables strict validation of an ABI-encoded prefix, allowing trailing bytes.
+    /// Allows trailing bytes when strict ABI decoding is enabled.
     ///
-    /// This implies [`strict`](Self::strict) and [`validate`](Self::validate), but skips
-    /// the final check that the encoding consumes the entire input. All other checks,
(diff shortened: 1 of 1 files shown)


crates/sol-types/src/abi/decoder.rs
```diff
@@ -89,13 +92,23 @@ impl AbiDecoderConfig {
...
 
     /// Returns whether strict ABI encoding is required.
+    ///
+    /// Trailing bytes are permitted when
+    /// [`get_validate_allow_trailing_bytes`](Self::get_validate_allow_trailing_bytes)
+    /// is enabled.
     #[inline]
     pub const fn get_strict(&self) -> bool {
-        self.strict
+        self.strict || self.validate_allow_trailing_bytes
```
DaniPopes: dont think we want this

klkvr:
```suggestion
        self.strict
```

DaniPopes (approved):

